PR #13, opened 2021-06-07
change the command API so as to not lump aggregate and event serialization into a single trait/argument

ckmcd:
This is required for analogous changes in surge-ukg

### DIFF
--- a/modules/common/src/main/scala/surge/core/SurgeFormatting.scala
+++ b/modules/common/src/main/scala/surge/core/SurgeFormatting.scala
@@ -13,7 +13,8 @@ trait SurgeAggregateReadFormatting[Agg] {
   def readState(bytes: Array[Byte]): Option[Agg]
 }
 
-//trait SurgeReadFormatting[Agg, Event] extends SurgeEventReadFormatting[Event] with SurgeAggregateReadFormatting[Agg]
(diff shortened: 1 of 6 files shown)


modules/common/src/main/scala/surge/core/SurgeFormatting.scala
```diff
@@ -22,6 +23,8 @@ trait SurgeEventWriteFormatting[Event] {
 trait SurgeAggregateWriteFormatting[Agg] {
   def writeState(agg: Agg): SerializedAggregate
 }
+
+@deprecated("Encourages the use of an anti-pattern. Use SurgeAggregateWriteFormatting and SurgeEventWriteFormatting separately.", "0.5.4")
```
jeffboutotte, line 27: Should we also deprecate `SurgeAggregateFormatting` and `SurgeEventFormatting` for the same reason?

jeffboutotte (approved):
LGTM

modules/common/src/main/scala/surge/core/SurgeFormatting.scala
ckmcd: No, I think it makes sense to have something that knows how to serde an aggregate or an event but not something that serializes aggregates and events and something else that deserializes them.

PR #14, opened 2021-06-09
add a sbt publishTo setting

ckmcd:


jeffboutotte (approved):
Seems reasonable until we actually cross that bridge - thanks!

PR #16, opened 2021-06-10
Use DSL specific types for contexts.

ckmcd:
Create per DSL Surge Context objects wrapping the core Context object.  Although we don't use the context yet, I anticipate that it will want a DSL specific interface.  Introducing this now will avoid breaking the API later.


modules/command-engine/core/src/main/scala/surge/core/Context.scala
```diff
@@ -2,11 +2,13 @@
...
 package surge.core
 
+import akka.actor.ActorRef
+
 import scala.concurrent.ExecutionContext
-import scala.concurrent.ExecutionContext.global
 
-object Context {
-  val noop: Context = new Context()
+private[surge] object Context {
```
jeffboutotte, line 9: If these are both going to be private now (which makes sense to me for the core one) we can move them to the `surge.internal` package

modules/command-engine/javadsl/src/main/scala/surge/javadsl/common/Context.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import surge.core.{ Context => Core }
+
+trait Context {}
+
+object Context {
+  def apply(core: Core): Context = new ContextImpl(core)
+}
+
+class ContextImpl(private val core: Core) extends Context
```
jeffboutotte, line 13: Minor but I think `ContextImpl` can be internal and we can keep the `Context` trait as the public/backwards compatible interface

modules/command-engine/scaladsl/src/main/scala/surge/scaladsl/common/Context.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import surge.core.{ Context => Core }
+
+trait Context {}
+
+object Context {
+  def apply(core: Core): Context = new ContextImpl(core)
+}
+
+class ContextImpl(private val core: Core) extends Context
```
jeffboutotte, line 13: Same here as the javadsl - the impl can be internal

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -199,6 +199,8 @@ class PersistentActor[S, M, R, E](
     super.preStart()
   }
 
+  private def surgeContext() = Context(businessLogic.executionContext, self)
```
jeffboutotte, line 202: How would you see a team interacting with the `ActorRef` here? There's a lot it's doing under the hood and it's very internal at the moment so my only concern would be that we could break consumers using the ActorRef relatively easily without knowing

ckmcd: I anticipate that we will expose functions that interact with actor, but I do not intend to expose the actor ref directly to the app.

For instance, something like

```
def scheduleAlarm(time: Instance) = ScheduleAlarm(time) ? actorRef
```

jeffboutotte (approved):

PR #18, opened 2021-06-15
Fix false serialization warning on metrics map passed around by StreamManager

ckmcd:
When passing around metrics wrap the map in a class annotated with NoSerializationVerificationNeeded.


modules/common/src/main/scala/surge/internal/streams/StreamManager.scala
```diff
@@ -107,7 +114,7 @@ class KafkaStreamManager[Key, Value](
 
   def getMetrics: Future[Map[MetricName, Metric]] = {
     implicit val timeout: Timeout = Timeout(metricFetchTimeout)
-    (managerActor ? KafkaStreamManagerActor.GetMetrics).mapTo[Map[MetricName, Metric]]
+    (managerActor ? KafkaStreamManagerActor.GetMetrics).mapTo[MetricsWrapper].map(_.metrics)(global)
```
jeffboutotte, line 117: Does `handleGetMetrics` below in the actor also need to return a `MetricsWrapper` object? The `mapTo` here might figure it out, but if the actor still sends back a `Map` won't the serialization fail the same way?

jeffboutotte (approved):

PR #24, opened 2021-06-23
wrap arrayblockingqueue in source

zachary-rote:


modules/common/src/main/scala/surge/internal/health/windows/stream/sliding/SlidingHealthSignalStream.scala
```diff
@@ -78,28 +79,25 @@ private class SlidingHealthSignalStreamImpl(
...
-      .newCachedThreadPool()
-      .submit(new Callable[Unit] {
-        override def call: Unit = {
-          while (true) {
-            val signal = signals.take()
-            windowActors.foreach(a => a.processSignal(signal))
-          }
-        }
-      })
+    val blockingSource = Source.unfoldResource[HealthSignal, ArrayBlockingQueue[HealthSignal]](() => signals, signals => Option(signals.take()), _ => {})
```
jeffboutotte, line 90: This seems like a bit of an awkward fit here since the `unfoldResource` is really meant for wrapping bounded things, like querying a bunch of records from a DB. I think maybe a `Source.queue` would be a better fit - we could bound it to something decently large so we likely won't drop elements but wouldn't grow unbounded. It may require a bit of plumbing to the superclass that feeds signals in. I think that should also eliminate the need for a RestartSource here.

modules/common/src/test/scala/surge/internal/health/windows/stream/sliding/SlidingHealthSignalStreamSpec.scala
```diff
@@ -56,7 +57,10 @@ class SlidingHealthSignalStreamSpec
           SignalNameEqualsMatcher("test.trace", Some(SideEffect(Seq(HealthSignal("health.signal", "boom", SignalType.ERROR, Error("bah", None)))))),
         Some(SideEffect(Seq(signal)))))
     val signalStreamProvider = new SlidingHealthSignalStreamProvider(
-      WindowingStreamConfig(maxDelay = FiniteDuration(1, "second"), advancerConfig = WindowingStreamSliderConfig()),
+      WindowingStreamConfig(
+        maxDelay = FiniteDuration(1, "second"),
```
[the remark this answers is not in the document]
zachary-rote: done

modules/common/src/main/scala/surge/internal/health/windows/stream/sliding/SlidingHealthSignalStream.scala
zachary-rote: looking into it.

modules/common/src/main/scala/surge/health/config/WindowingStreamConfig.scala
```diff
@@ -51,9 +54,18 @@ object WindowingStreamConfigLoader {
     val maxStreamSize = config.getInt("surge.health.window.stream.max-size")
     val frequencies = config.getDurationList("surge.health.window.stream.frequencies").asScala.map(d => FiniteDuration(d.toMillis, "millis"))
 
+    val drainPerTick = Try { config.getInt("surge.health.stream.drain-per-tick") }.toOption.getOrElse(10)
```
jeffboutotte, line 57: No need to wrap these in a `Try` - we should always have default values defined for these so it shouldn't fail. If someone horribly misconfigures it also may be better to explode than quietly ignore their setting & use a default

modules/common/src/main/scala/surge/health/config/WindowingStreamConfig.scala
```diff
@@ -51,9 +54,18 @@ object WindowingStreamConfigLoader {
...
+
+    val tickInterval =
+      Try { config.getDuration("surge.health.stream.tick-duration") }.toOption.map(d => FiniteDuration(d.toMillis, "millis")).getOrElse(250.millis)
+
+    val initialTickDelay =
+      Try { config.getDuration("surge.health.stream.initial-tick-delay") }.toOption.map(d => FiniteDuration(d.toMillis, "millis")).getOrElse(1.second)
```
jeffboutotte, line 63: Same here with `FiniteDuration` - just do a `config.getDuration("surge.health.stream.initial-tick-delay").toMillis.millis` in case of stray typos

modules/common/src/main/scala/surge/internal/health/windows/stream/WindowingHealthSignalStream.scala
```diff
@@ -2,41 +2,55 @@
...
    */
-  override def signalHandler: SignalHandler = (signal: HealthSignal) => {
-    // todo: consider ways to handle failed offer to add signal to blocking queue
+  final override def signalHandler: SignalHandler = (signal: HealthSignal) => {
     Try {
-      if (!signals().add(signal)) {
-        throw new RuntimeException("failed to add signal. need too handle this case")
-      }
+      sourceQueue().foreach(queue => {
+        Source.single(signal).runWith(Sink.foreach(s => queue.offer(s)))(Materializer(actorSystem))
```
jeffboutotte, line 52: Does just `queue.offer(signal)` work here?

modules/common/src/main/scala/surge/internal/health/windows/stream/sliding/SlidingHealthSignalStream.scala
```diff
@@ -77,31 +88,30 @@ private class SlidingHealthSignalStreamImpl(
...
 
     // Stream Handle
-    () => {
-      future.cancel(true)
-      windowActors.foreach(actor => actor.stop())
-    }
+    new SlidingHealthSignalStreamHandle(
+      windowActors,
+      () => {
+        signalData = null
```
jeffboutotte, line 112: I'd avoid this null - if it's something that may or may not exist make it an explicit `Option` so consumers are forced to deal with the case it's not there. If it exists as a non-optional type coming from scala code it's generally assumed that it's not null.

modules/common/src/main/scala/surge/internal/health/windows/stream/WindowingHealthSignalStream.scala
```diff
@@ -2,41 +2,55 @@
...
 import surge.health.domain.HealthSignal
-import surge.health.matchers.SignalPatternMatcher
-import surge.health.windows.Window
+import surge.utils.BooleanWrapper
 
 import scala.util.Try
 
 trait StreamHandle {
-  def release(): Unit
+  private val running: BooleanWrapper = new BooleanWrapper().asTrue()
```
jeffboutotte, line 16: Hmm let's just make this a `private var` for now and avoid the `BooleanWrapper` altogether. It's still mutable under the hood using a mutable map so it doesn't really buy us much and makes it a bit more abstract

modules/common/src/main/scala/surge/internal/health/windows/stream/sliding/SlidingHealthSignalStream.scala
zachary-rote: done

modules/common/src/main/scala/surge/internal/health/windows/stream/WindowingHealthSignalStream.scala
zachary-rote: yes

modules/common/src/main/scala/surge/internal/health/windows/stream/sliding/SlidingHealthSignalStream.scala
zachary-rote: yeah. this whole var vs val thing is a bit disconcerting.  Especially when the style guide always `var shames` me.  lol

modules/common/src/main/scala/surge/internal/health/windows/stream/WindowingHealthSignalStream.scala
zachary-rote: point taken.  again trying to avoid the `var shaming`..

modules/common/src/main/scala/surge/health/config/WindowingStreamConfig.scala
zachary-rote: done

modules/common/src/main/scala/surge/health/config/WindowingStreamConfig.scala
zachary-rote: done

modules/common/src/main/scala/surge/internal/health/windows/stream/sliding/SlidingHealthSignalStream.scala
zachary-rote: done

modules/common/src/main/scala/surge/internal/health/windows/stream/WindowingHealthSignalStream.scala
zachary-rote: done

modules/common/src/main/scala/surge/kafka/KafkaProducer.scala
```diff
@@ -127,10 +127,8 @@ trait KafkaProducerTrait[K, V] extends KafkaSecurityConfiguration with KafkaProd
     try {
       producer.initTransactions()
     } catch {
-      case e: Exception => {
-        println(e)
+      case e: Exception =>
```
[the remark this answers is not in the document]
zachary-rote: I was debugging something and it fell through the cracks in a prior PR.

garyKeorkunian (approved):

PR #25, opened 2021-06-23
Simplify KStream Indexing Process

jeffboutotte:
For the indexing that happens on the consuming side of Surge using KStreams to populate a KTable of aggregate states we had a concept of a "validation processor" which would allow an app to raise warnings based on comparing an old aggregate state to the newly discovered one (essentially determine if the state transition was an expected on on the KStreams side of things).  In theory and when initially testing the framework this is useful for continuously validating we're not seeing any corruption in the event stream - for example using sequence numbers to determine if something went backwards in time. In practice, however, this hasn't really proved to be all that useful since we've hardened the framework to the point where we never see any of these warnings based on app validation.

Even though this feature is theoretically useful, in order to grab the previous state for comparison and not need a changelog topic & duplicate data specifically for that we were extending a class internal to Kafka which pretty tightly couples Surge to the particular version of Kafka that we use rather than being flexible.  Because we're not really leveraging this anyways, this trade-off is really not something that's worth it given the fragility it introduces.


modules/common/src/test/scala/surge/internal/health/supervisor/HealthSupervisorActorSpec.scala
```diff
@@ -51,8 +51,8 @@ class HealthSupervisorActorSpec
       bus.signalStream().start()
 
       // Register
-      val done = bus.register(probe.ref, componentName = "boomControl", Seq(Pattern.compile("boom")))
-      done shouldBe a[Future[Ack]]
+      val done = Await.result(bus.register(probe.ref, componentName = "boomControl", Seq(Pattern.compile("boom"))), 10.seconds)
```
sebastianharko, line 54: you could mix in the ScalaFutures trait from ScalaTest (http://doc.scalatest.org/3.0.1-2.12/org/scalatest/concurrent/ScalaFutures.html) to avoid the Await.result ... up to you though ;-)

jeffboutotte: Great idea! Will do.

sebastianharko:
> this is useful for continuously validating we're not seeing any corruption in the event stream - for example using sequence numbers to determine if something went backwards in time. In practice,

I like this idea and it could also be implemented as some sort of micro-service that sits _outside_ the Surge apps and scans Kafka Topics used by Surge apps for corruption. What do you think ?

jeffboutotte:
That makes a lot of sense to me! We could add a small utility to help with that sort of thing. It could be embedded alongside the command app (self validating in the background) or could be added to a completely separate micro-service outside of the command app specifically used for validation

PR #26, opened 2021-06-25
Remove the separate thread pool for the aggregate actors and add a small utility to ...

sebastianharko:
Remove separate thread pool for the aggregate actors and add a small utility to detect blocked execution contexts


modules/common/src/main/resources/reference.conf
```diff
@@ -280,3 +280,17 @@ surge {
   dr-standby-enabled = false
   dr-standby-enabled = ${?SURGE_ENABLE_DR_STANDBY}
 }
+
+prober {
```
jeffboutotte, line 284: Maybe `execution-context-prober` or something on the naming here?

modules/common/src/main/scala/surge/internal/utils/ExecutionContextProber.scala
```diff
@@ -0,0 +1,155 @@
...
+  def collect(count: Int, expectedId: UUID): Receive = {
+    case Timeout =>
+      log.warning(s"One of our probes timed out. ${warningText}.")
+      // the execution context is potentially unhealthy so we want to suspend our checks for a bit (to give
+      // it time to recover in case it's just a hiccup), hence interval * 3.
+      detectedIssue = true
+      timers.startSingleTimer(SendProbesKey, msg = SendProbes, settings.interval * 3)
+      context.become(receive)
+    case Status.Failure(e) =>
+      log.error(e, s"One of our probes resulted in a failed future. ${warningText}.")
```
jeffboutotte, line 98: It'd be good to add some details about which execution context in particular we're using to the couple of logs here. Right now It will just be the default for the actor system but I can us reusing this for other execution contexts as well.

modules/common/src/main/scala/surge/internal/utils/ExecutionContextProber.scala
```diff
@@ -0,0 +1,155 @@
...
+}
+
+class ExecutionContextProberActor(settings: ExecutionContextProberSettings) extends Actor with ActorLogging with Timers {
+
+  import ExecutionContextProberActor.Messages._
+  import ExecutionContextProberActor._
+
+  implicit val system: ActorSystem = context.system
+
+  var detectedIssue = false // used by the unit test
```
jeffboutotte, line 67: I'd keep this private since it's pretty internal to the actor

jeffboutotte (commented):
Looking forward to this change! Just a couple minor things

modules/common/src/main/resources/reference.conf
sebastianharko: Yes, you are right execution-context-prober sounds better ;-) I'm going to make the change.

modules/common/src/main/scala/surge/internal/utils/ExecutionContextProber.scala
sebastianharko: True, I'll add some more details

modules/common/src/main/scala/surge/internal/utils/ExecutionContextProber.scala
sebastianharko: Yes, so this is interesting, because in idiomatic Akka "private" doesn't make much sense, since all internal state is private anyway. So, in other words:

```
val myActor: ActorRef = system.actorOf(Props(new MyActor()), name = "myActor") 
myActor.detectedIssues // this won't work anyway - it doesn't make a difference if it's private or public
```

modules/common/src/main/scala/surge/internal/utils/ExecutionContextProber.scala
sebastianharko: Done, marking some resolved

modules/common/src/main/resources/reference.conf
```diff
@@ -280,3 +280,17 @@ surge {
...
+    initial-delay = 3 seconds
+    timeout = 50 milliseconds
+    interval = 5 seconds
+    num-probes = 7
+    dispatcher {
```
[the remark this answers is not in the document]
sebastianharko: Yeah sorry mea culpa ... The prober should run on its own dispatcher (not only in tests but also in prod) so it can be immune from thread starvation issues ... So, it needs to sit 'outside' so it can examine what's going on with the other execution contexts , let me know if it makes sense ;-)

modules/common/src/main/scala/surge/internal/utils/ExecutionContextProber.scala
sebastianharko: @garyKeorkunian Fixed (now the variable is inside the receive context) ;-)

garyKeorkunian (approved):

PR #29, opened 2021-07-06
Update sequence diagram

garyKeorkunian:
@sebastianharko , here is the sequence diagram.  There was one already there, but this update has some more detail.


garyKeorkunian:
Updated the PR to add the diagram to the surge-docs module.

jeffboutotte (approved):

sebastianharko (approved):

PR #31, opened 2021-07-06
Propagate MDC between actors

jeffboutotte:


jeffboutotte:
@sebastianharko as you're digging into some of the OpenTracing pieces, some of the work on this branch may be useful. The work here was to propagate the MDC between actors/across the network as needed, but it's all plumbed through alongside the OpenTracing bits. There were a couple places that `TracedMessages` were unwrapped by an actor but never carried through to outgoing messages, so it's likely we were losing spans along the way. I think I've addressed at least a couple of those places, so while this work may not be completely ready it's likely you'll run into & fix a couple of the same things things I have on this branch.  Feel free to grab whatever code from here is useful as you dig in!

sebastianharko:
@jeffboutotte Awesome, I like, I'll have a look ;-)

garyKeorkunian (approved):

PR #41, opened 2021-07-12
refactor event sinks to support metrics and better logging

ckmcd:


modules/common/src/main/scala/surge/streams/EventSink.scala
```diff
@@ -28,3 +36,11 @@ trait EventSink[Event] extends EventHandler[Event] {
     FlowConverter.flowFor(handleEvent, partitionBy, sinkExceptionHandler, parallelism)(ExecutionContext.global)
   }
 }
+
+trait EventSinkSupport[Event] {
+  def metrics: Metrics = Metrics.globalMetricRegistry
+  def metricTags: Map[String, String] = Map("class" -> getClass.getCanonicalName)
+  def formatMetaForExceptionHandlerMessage[Meta]: EventPlusStreamMeta[String, Event, Meta] => String = _.streamMeta.toString
```
jeffboutotte, line 43: It seems like we could just add these directly to `EventHandler`. Metrics stuff seems like it should be available for any `EventHandler` as we add more observability going forward and an `EventHandler` that does not extend `EventSinkSupport` would get the same implementation as this default anyways via the `DefaultDataSinkExceptionHandler`.

modules/common/src/test/scala/surge/internal/health/matchers/RepeatingSignalMatcherSpec.scala
```diff
@@ -63,8 +63,8 @@ class RepeatingSignalMatcherSpec extends TestKit(ActorSystem("RepeatingSignals")
         Range(1, 100),
         Range(1, 100))
 
-      Source(repeatingData)
-        .mapAsync(parallelism = 10)(data =>
+      Source(repeatingData.asInstanceOf[immutable.Iterable[Range]])
```
jeffboutotte, line 66: Cleaning up a warning here I'm guessing? Maybe following the pattern in the Akka Streams docs for [`Source.from`](https://doc.akka.io/docs/akka/current/stream/operators/Source/from.html) would be better here? Even though this is just tests, `asInstanceOf` makes me nervous in general

jeffboutotte (commented):
Awesome ideas, making things more observable is always better!

modules/common/src/main/scala/surge/streams/EventSink.scala
jeffboutotte: Chatted live on this and we'll keep it the way it is to preserve backwards compatibility for now at least. We can look at if it makes sense to revisit later

ckmcd:
No longer needed.

PR #42, opened 2021-07-13
Tracing

sebastianharko:


modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -165,13 +163,7 @@ class PersistentActor[S, M, R, E](
 
   private case class EventPublishTimedOut(reason: Throwable, startTime: Instant) extends Internal
 
-  protected case class InternalSpans(messageProcessingSpan: Option[Span] = None)
```
sebastianharko, line 168: removing these internal spans from here for now because they were getting in my way

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -198,6 +190,11 @@ class PersistentActor[S, M, R, E](
   private val publishStateOnly: Boolean = businessLogic.kafka.eventsTopicOpt.isEmpty
 
   assert(!publishStateOnly || businessLogic.eventWriteFormattingOpt.nonEmpty, "businessLogic.eventWriteFormattingOpt may not be none when publishing events")
+
+  override def messageNameForTracedMessages: MessageNameExtractor = { case t: ProcessMessage[_] =>
+    s"ProcessMessage[${t.message.getClass.getSimpleName}]"
+  }
```
sebastianharko, line 196: ActorWithTracing now provides users the opportunity to provide a "pretty name" for actor messages via this partial function. If a nice name is not defined for a given message, we default .getClass.getSimpleName (which doesn't always work well ...)

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -225,9 +223,8 @@ class PersistentActor[S, M, R, E](
   }
 
   private def handle(state: InternalActorState, msg: ProcessMessage[M]): Unit = {
-    val ProcessMessageSpan = createSpan("process_message").setTag("aggregateId", aggregateId)
     context.setReceiveTimeout(Duration.Inf)
-    context.become(persistingEvents(state.withMessageProcessingSpan(ProcessMessageSpan)))
```
sebastianharko, line 230: got rid of .withMessageProcessingSpan for now because it was getting in my way, but I'll get back to adding more details wrt to PersistentActor in a subsequent PR

modules/common/src/main/scala/surge/kafka/KafkaPartitionShardRouterActor.scala
```diff
@@ -161,43 +158,36 @@ class KafkaPartitionShardRouterActor(
...
-        log.trace(
-          s"RouterActor forwarding command envelope for aggregate $aggregateId to region ${responsiblePartitionRegion.regionManager.pathString}. Msg $msg")
-
-        responsiblePartitionRegion.regionManager.forward(msg)
+        log.trace(s"Forwarding command envelope for aggregate $aggregateId to region ${responsiblePartitionRegion.regionManager.pathString}.")
+        val tracedMsg = TracedMessage(msg, parentSpan = activeSpan)
+        responsiblePartitionRegion.regionManager.forward(tracedMsg)
       case None =>
-        log.error(s"RouterActor could not find a responsible partition region for $aggregateId.")
+        log.error(s"Could not find a responsible partition region for $aggregateId.")
```
sebastianharko, line 190: the log message doesn't need the name of the actor because the name of the actor (or at least the class) should already be shown by logback ...

modules/common/src/main/scala/surge/tracing/TracedMessage.scala
```diff
@@ -2,45 +2,18 @@
...
 
-case class TracedMessage[T](
+final case class TracedMessage[T](
     @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "messageType", visible = true) message: T,
     headers: Map[String, String])
-    extends JacksonSerializable {
-  private def spanContext(tracer: Tracer): Option[SpanContext] = {
-    Option(
-      tracer.extract(
-        Format.Builtin.TEXT_MAP_EXTRACT,
```
sebastianharko, line 35: this wasn't working with Jaeger ;-( 
Jaeger only seems to like TEXT_MAP and not TEXT_MAP_EXTRACT AND TEXT_MAP_INJECT

modules/surge-docs/src/test/scala/docs/command/BankAccountSurgeModelWithTracer.scala
```diff
@@ -0,0 +1,59 @@
...
+import io.opentracing.Tracer
+import play.api.libs.json.Json
+import surge.core._
+import surge.kafka.KafkaTopic
+import surge.scaladsl.command.{ AggregateCommandModel, SurgeCommandBusinessLogic }
+
+import java.util.UUID
+
+// #surge_model_class
+// format: off
```
sebastianharko, line 16: didn't look good with scalafmt ;-(

build.sbt
```diff
@@ -1,5 +1,6 @@
 // Copyright © 2017-2020 UKG Inc. <https://www.ukg.com>
 
+import Dependencies.autoImport.OpenTracing.jaeger
```
sebastianharko, line 3: surge docs now contains some example code that shows how to initialize a jaeger tracer so that's why we need this here

modules/common/src/main/scala/surge/internal/akka/ActorWithTracing.scala
```diff
@@ -3,29 +3,75 @@
...
-import io.opentracing.Scope
-import surge.internal.utils.SpanSupport
-import surge.tracing.TracedMessage
+import akka.actor.ActorRef
+import io.opentracing._
+import surge.tracing.{ SpanExtensions, TracePropagation, TracedMessage }
 
-trait ActorWithTracing extends AroundReceiveActor with SpanSupport {
-  private var activeScope: Option[Scope] = None
+trait ActorWithTracing extends AroundReceiveActor with ActorOps with SpanExtensions {
```
sebastianharko, line 10: One change is that every message gets a span called "ActorName:Message" (i.e. "PersistentActor:ProcessCommand[AddItem]" )

When the message is a TraceableMessage the span gets a parent span. Otherwise, the span doesn't have a parent. 

Can this lead to too many spans ? Yes - but we can use the sampler configuration to discard spans we don't need.

modules/common/src/main/scala/surge/tracing/TracedMessage.scala
```diff
@@ -2,45 +2,18 @@
...
-        new TextMapExtract {
-          override def iterator(): util.Iterator[util.Map.Entry[String, String]] = headers.asJava.entrySet().iterator()
-        }))
-  }
-
-  def activeSpan(tracer: Tracer): Span = {
```
sebastianharko, line 41: Refactored and move some of this code to a object called TracePropagation (whose only purpose is to serialize/deserialize headers)

modules/common/src/main/scala/surge/tracing/TracedMessage.scala
jeffboutotte: Interesting does it work with Zipkin as well? I'm good with whatever way it works as long as it works across the board :)

version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "0.5.15-SNAPSHOT"
+ThisBuild / version := "0.5.16-SNAPSHOT"
```
jeffboutotte, line 1: No need to bump the snapshot version here, the version just gets bumped as part of the release process

modules/common/src/main/scala/surge/internal/akka/ActorWithTracing.scala
```diff
@@ -3,29 +3,75 @@
...
+        val operationName: String = s"$actorClassSimpleName:$messageName"
+        val newSpan: Span = TracePropagation.childFrom(tracedMsg, operationName)
+        activeSpan = newSpan
+        val fields: Map[String, String] = getFields +
+          ("message" -> messageName) +
+          ("message (fqcn)" -> tracedMsg.message.getClass.getName)
+        newSpan.log("receive", fields)
+        superAroundReceive(receive, tracedMsg.message)
+      case other =>
+        val messageName: String = other.getClass.getSimpleName
```
jeffboutotte, line 53: Any reason not to do a `getMessageName` here as well?

jeffboutotte (commented):
Looking good! Just a few really minor things

modules/common/src/main/scala/surge/tracing/TracedMessage.scala
sebastianharko: Yes.. TEXT_MAP seems to work with both zipkin and jaeger ;-)

modules/common/src/main/scala/surge/internal/akka/ActorWithTracing.scala
sebastianharko: good point, not a bad idea to do a getMessageName even for messages that are not TracedMessages ...

build.sbt
```diff
@@ -96,7 +97,7 @@ lazy val `surge-docs` = (project in file("modules/surge-docs"))
   .settings(
     skip in publish := true,
     paradoxTheme := Some(builtinParadoxTheme("generic")),
-    libraryDependencies ++= Seq(typesafeConfig, embeddedKafka, logback, scalatest, scalatestPlusMockito, mockitoCore))
+    libraryDependencies ++= Seq(typesafeConfig, embeddedKafka, logback, scalatest, scalatestPlusMockito, mockitoCore, jaeger))
```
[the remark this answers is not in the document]
sebastianharko: Done

version.sbt
sebastianharko: Fixed

modules/common/src/main/scala/surge/internal/akka/ActorWithTracing.scala
sebastianharko: done ;-) also removed some duplication

modules/common/src/main/scala/surge/tracing/SpanHelper.scala
```diff
@@ -1,6 +1,6 @@
 // Copyright © 2017-2021 UKG Inc. <https://www.ukg.com>
 
-package surge.internal.utils
+package surge.tracing
```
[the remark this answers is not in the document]
sebastianharko: @jeffboutotte Done, moved the tracing stuff to an _internal_ package ...

PR #46, opened 2021-07-20
Feature/enable disable health signal stream

zachary-rote:


modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -92,64 +129,70 @@ class KafkaProducerActor(
...
     }
   }
 
-  override def stop(): Unit = {
-    publisherActor ! ActorLifecycleManagerActor.Stop
+  override def start(): Future[Ack] = {
+    implicit val askTimeout: Timeout = Timeout(TimeoutConfig.PublisherActor.askTimeout)
+
+    val result = publisherActor
+      .ask(ActorLifecycleManagerActor.Start)
```
sebastianharko, line 149: use mapTo[ActorLifecycleManagerActor.Ack]  instead of map and you won't have to worry about case _ and throwing an exception

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -131,56 +131,31 @@ class KafkaProducerActor(
...
   override def stop(): Future[Ack] = {
     implicit val askTimeout: Timeout = Timeout(TimeoutConfig.LifecycleManagerActor.askTimeout)
 
-    val result = publisherActor.ask(ActorLifecycleManagerActor.Stop).map[Ack] {
-      case _: ActorLifecycleManagerActor.Ack =>
-        Ack()
-      case _ =>
-        throw new RuntimeException("Unexpected response from actor stop request")
-    }
+    val result = publisherActor.ask(ActorLifecycleManagerActor.Stop).mapTo[ActorLifecycleManagerActor.Ack].map(_ => Ack())
```
jeffboutotte, line 152: I'd just do `publisherActor.ask(ActorLifecycleManagerActor.Stop).map(_ => Ack())` here since we don't actually use the value we get back from the `ActorLifecycleManagerActor`

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -131,56 +131,31 @@ class KafkaProducerActor(
...
-    val result = publisherActor
-      .ask(ActorLifecycleManagerActor.Start)
-      .map {
-        case _: ActorLifecycleManagerActor.Ack =>
-          Ack()
-        case _ =>
-          throw new RuntimeException("Unexpected response from actor start request")
-      }
-      .andThen(registrationCallback())
+    val result = publisherActor.ask(ActorLifecycleManagerActor.Start).mapTo[ActorLifecycleManagerActor.Ack].map(_ => Ack()).andThen(registrationCallback())
```
jeffboutotte, line 144: Same here - no need to `mapTo` since we never use the value

modules/common/src/main/scala/surge/internal/akka/actor/ActorLifecycleManagerActor.scala
```diff
@@ -8,7 +8,7 @@ import surge.internal.akka.actor.ActorLifecycleManagerActor.Ack
 object ActorLifecycleManagerActor {
   case object Start
   case object Stop
-  case class Ack(success: Boolean)
+  case class Ack()
```
jeffboutotte, line 11: I think we can just leverage the common `Ack()` here rather than a custom `ActorLifecycleManagerActor.Ack()` since they're identical now

modules/common/src/main/scala/surge/internal/health/HealthSignalStreamProvider.scala
```diff
@@ -85,7 +89,37 @@ trait HealthSignalStreamProvider {
...
-  def bus(): HealthSignalBusInternal = {
-    signalBus
+  def bus(): HealthSignalBusInternal = signalBus
+}
+
+class NullHealthSignalStream(config: HealthSignalBusConfig, bus: HealthSignalBusTrait, override val actorSystem: ActorSystem) extends HealthSignalStream {
+
+  override def signalHandler: SignalHandler = (_: HealthSignal) =>
+    Try {
+      Done
```
jeffboutotte, line 99: Slight optimization you can do `Success(Done)` here

modules/common/src/test/scala/surge/internal/health/windows/actor/HealthSignalWindowActorSpec.scala
```diff
@@ -17,11 +17,11 @@ import surge.internal.health.windows.WindowSlider
 import surge.internal.health.{ HealthSignalBus, HealthSignalBusInternal }
 
 import scala.concurrent.duration._
-import scala.languageFeature.postfixOps
 class HealthSignalWindowActorSpec
     extends TestKit(ActorSystem("HealthSignalWindowActorSpec", ConfigFactory.load("artery-test-config")))
     with AnyWordSpecLike
     with Matchers
+    with BeforeAndAfterEach
```
jeffboutotte, line 24: Doesn't look like this is used at the moment

project/Dependencies.scala
```diff
@@ -37,8 +37,8 @@ object Dependencies extends AutoPlugin {
 
       val version = "1.4.1"
       val api = "io.opentelemetry" % "opentelemetry-api" % version
-      val sdk = "io.opentelemetry" % "opentelemetry-sdk" % OpenTelemetry.version % Test
-      val sdkTesting = "io.opentelemetry" % "opentelemetry-sdk-testing" % OpenTelemetry.version % Test
+      val sdk = "io.opentelemetry" % "opentelemetry-sdk" % OpenTelemetry.version
+      val sdkTesting = "io.opentelemetry" % "opentelemetry-sdk-testing" % OpenTelemetry.version
```
jeffboutotte, line 41: These might be merge conflicts of some kind? the opentelemetry-sdk stuff should be just test dependencies though - libraries should only depend on the API.

project/Dependencies.scala
```diff
@@ -72,4 +72,8 @@ object Dependencies extends AutoPlugin {
     val slf4jApi = "org.slf4j" % "slf4j-api" % "1.7.30"
     val typesafeConfig = "com.typesafe" % "config" % "1.4.1"
   }
+
+  object Jackson {
+    val databind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.10.1"
```
jeffboutotte, line 77: Is this used anywhere? I think Akka pulls through a Jackson dependency for remote serialization but I'd rather not explicitly depend on it unless we really need

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
zachary-rote: done

modules/common/src/main/scala/surge/internal/health/HealthSignalStreamProvider.scala
zachary-rote: done

modules/common/src/test/scala/surge/internal/health/windows/actor/HealthSignalWindowActorSpec.scala
zachary-rote: done

project/Dependencies.scala
zachary-rote: ActorWithTracing depends on `io.opentelemetry.api.trace.{ Span, Tracer }`

project/Dependencies.scala
zachary-rote: getting build failures without it..

project/Dependencies.scala
jeffboutotte: I think those both come from the `opentelemetry-api` dependency

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
zachary-rote: done

modules/common/src/main/scala/surge/internal/akka/actor/ActorLifecycleManagerActor.scala
zachary-rote: done

project/Dependencies.scala
zachary-rote: done

project/Dependencies.scala
zachary-rote: done

jeffboutotte (approved):
LGTM

jeffboutotte:
You may want to rename the PR title of this as well so it's not merged with "DRAFT" in it :)

PR #47, opened 2021-07-21
Internal Logging & Test Cleanup

jeffboutotte:
Lots of internal cleanup including:
- Kafka consumer group SubscriptionInfo class now gracefully handles cases where the bytes received are from a non-surge consumer & different than expected.
- KafkaSecurityConfiguration is available as an object that can be called statically in addition to a trait that can be mixed in.
- Remove a few unnecessary calls to `asInstanceOf`
- Properly shut down the test actor system created in tests using Akka TestKit.
- Drop the logging level for a couple of internal health related components to trace since they're a bit noisy even for debug level.
- Allow configuring tick frequency of `HealthSignalWindowActor` & make tests for that actor faster
- Reduce timing settings & patience config in `HealthSupervisorActorSpec` so those tests run faster
- Fix a race condition in `KafkaProducerActorImplSpec` when expecting both a failure response and actor termination at the same time.


modules/common/src/test/scala/surge/internal/akka/streams/graph/OptionFlowSpec.scala
```diff
@@ -6,10 +6,16 @@ import akka.actor.ActorSystem
...
+import org.scalatest.BeforeAndAfterAll
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpecLike
 
-class OptionFlowSpec extends TestKit(ActorSystem("OptionFlowSpec")) with AnyWordSpecLike with Matchers {
+class OptionFlowSpec extends TestKit(ActorSystem("OptionFlowSpec")) with AnyWordSpecLike with Matchers with BeforeAndAfterAll {
+
+  override def afterAll(): Unit = {
+    TestKit.shutdownActorSystem(system)
+    super.afterAll()
```
sebastianharko, line 17: I don't think that super.afterAll is required since in this case there's no other trait that uses BeforeAndAfterAll ... The afterAll in the BeforeAndAfterAll trait does nothing (is empty)

modules/command-engine/core/src/test/scala/surge/core/AggregateRefTraitSpec.scala
```diff
@@ -15,7 +16,17 @@ import surge.internal.persistence.{ AggregateRefTrait, PersistentActor }
...
-class AggregateRefTraitSpec extends TestKit(ActorSystem("AggregateRefTraitSpec")) with AsyncWordSpecLike with Matchers with ScalaFutures {
+class AggregateRefTraitSpec
+    extends TestKit(ActorSystem("AggregateRefTraitSpec"))
+    with AsyncWordSpecLike
+    with Matchers
+    with ScalaFutures
+    with BeforeAndAfterAll {
+
+  override def afterAll(): Unit = {
+    TestKit.shutdownActorSystem(system)
```
sebastianharko, line 27: Can we .shutdownActorSystem(system, verifySystemShutdown = true) so that we can get a more eye catching error ?

jeffboutotte: Yep - will do

modules/command-engine/core/src/test/scala/surge/internal/domain/SurgeMessagePipelineSpec.scala
```diff
@@ -25,7 +27,6 @@ import surge.internal.health.windows.stream.sliding.SlidingHealthSignalStreamPro
 import surge.kafka.streams.{ AggregateStateStoreKafkaStreams, MockPartitionTracker, MockState }
 import surge.metrics.Metrics
 
-import java.util.regex.Pattern
 import scala.concurrent.duration._
 
 @Ignore
```
jeffboutotte, line 32: Leaving this as ignored for now. Updating the timing has helped a few tests pass a little more reliably & quickly but some still fail with the shorter times.  I'm wondering if it's really just trying to do too many things & needs to be broken down a bit more to test smaller components.

On the few that do pass reliably, there's still an issue where something isn't shut down and the logs for the remainder of testing are flooded with errors connecting to Kafka (since the embedded brokers are spun down after the tests) which makes it pretty challenging to debug any other test failures.  I think before re-enabling this particular suite at least this should be figured out.

sebastianharko (approved):

PR #51, opened 2021-07-26
Instrument event sink

sebastianharko:


modules/common/src/main/scala/surge/streams/EventSink.scala
```diff
@@ -20,11 +22,29 @@ trait EventHandler[Event] {
...
+  def sinkName: String = this.getClass.getSimpleName
   def parallelism: Int = 8
   def handleEvent(key: String, event: Event, headers: Map[String, Array[Byte]]): Future[Any]
   def partitionBy(key: String, event: Event, headers: Map[String, Array[Byte]]): String
 
+  // execute the handleEvent function with tracing
+  private def execHandleEvent(key: String, event: Event, headers: Map[String, Array[Byte]]): Future[Any] = {
+    import surge.internal.tracing.TracingHelper._
+    val operationName = s"$sinkName:${event.getClass.getSimpleName}"
+    val span = tracer.spanBuilder(operationName).setNoParent().startSpan()
```
jeffboutotte, line 35: Is there a decent way to detect if this span follows from another one? I think it'd be more a follows_from than a parent child relationship in the source->sink case

This handler logic is ultimately hooked into an `EventSource` which provides the message reading/deserialization pieces of a stream processor. Eventually I think we'd want the span to trace all the way from there - for example, start when a message is read from Kafka, send to the sink here and continue the trace, additionally carry through the sink side for certain sinks we own (like the command engine sink since we've instrumented all the way through that)

modules/common/src/main/scala/surge/streams/EventSink.scala
```diff
@@ -20,11 +22,29 @@ trait EventHandler[Event] {
 }
 
 trait EventSink[Event] extends EventHandler[Event] {
+  val tracer: Tracer
```
jeffboutotte, line 25: For consistency with what we expect the user to provide on the command engine side we should make this an instance of OpenTelemetry and create the tracer ourselves like we do there

Additionally I think it'd be good to default to no-op so users don't need to think about it out of the box if they're not using tracing

sebastianharko: You are correct

modules/common/src/main/scala/surge/streams/EventSink.scala
sebastianharko: Yes, working on it

modules/common/src/main/scala/surge/internal/streams/StreamManager.scala
```diff
@@ -207,7 +221,7 @@ class KafkaStreamManagerActor[Key, Value](
     case _            => stash()
   }
 
-  private def startConsumer(): Unit = {
+  private def startConsumer(tracer: Tracer): Unit = {
```
jeffboutotte, line 224: Any reason in particular to pass the tracer to this function rather than leverage the one passed to the constructor? Looks like that's the one passed along anyways from the `Receive` function

modules/metrics/src/test/scala/surge/metrics/TimerSpec.scala
```diff
@@ -22,7 +22,7 @@ class TimerSpec extends MetricsSpecLike with ScalaFutures {
       metricValue(testTimerName) shouldEqual 95.5
     }
 
-    "Properly time scala Future completion time" in {
+    "Properly time scala Future completion time" ignore {
```
jeffboutotte, line 25: I think these can be un-ignored, they should be fixed (or more stable) with the latest commit on main.

modules/rabbit-support/src/main/scala/surge/rabbit/RabbitDataSource.scala
```diff
@@ -37,6 +37,8 @@ trait RabbitDataSource[Key, Value] extends DataSource[Key, Value] {
 
   def bufferSize: Int
 
+  val tracer: Tracer
```
jeffboutotte, line 40: This may want a default no-op - or may actually get a default no-op tracer from the `DataSource` parent trait already

modules/common/src/main/scala/surge/streams/KafkaEventSource.scala
```diff
@@ -35,34 +39,44 @@ trait EventSourceDeserialization[Event] {
...
+    val key = eventPlusOffset.messageKey
+    Try(eventDeserializationTimer.time(formatting.readEvent(payloadByteArray))) match {
+      case Failure(exception) =>
+        log.error("Unable to read event from byte array", exception)
+        span.log("failed to deserialize")
+        span.error(exception)
+        span.end()
+        Left(eventPlusOffset.streamMeta)
+      case Success(event) =>
+        span.log("deserialized", Map("event" -> event.getClass.getSimpleName))
```
jeffboutotte, line 57: Where does this span ultimately end in the success case here? Is that still something to be added at the offset commit stage since it looks like this span is passed from the subscription to Kafka? Additionally do we want to end the span in the error case here? Or have the offset commit stage at the very end of the stream end the span on either success or failure?

modules/common/src/main/scala/surge/internal/streams/StreamManager.scala
sebastianharko: @jeffboutotte I think you are correct, I'll fix it.

modules/metrics/src/test/scala/surge/metrics/TimerSpec.scala
sebastianharko: @jeffboutotte Awesome, I'll unignore.

modules/common/src/main/scala/surge/streams/KafkaEventSource.scala
sebastianharko: Yes, so I see three logical things here:
1. Fetching the message from Kafka
2. Deserializing the message and running the business logic
3. Offset commit stage

(1) We won't be able to capture in a span because it's deep in the Alpakka Kafka source so let's ignore it for now.
(2) It's what we capture now and the span ends upon completion of the execution of business logic
(3) I propose that the offset commit stage has its own span that "follows through" from (2).

jeffboutotte: Cool - that makes sense to me!

garyKeorkunian (approved):

PR #52, opened 2021-07-26
allow actorlifecyclemanager to accept stop msg when already stopped

zachary-rote:


modules/common/src/test/scala/surge/internal/health/windows/actor/HealthSignalWindowActorSpec.scala
```diff
@@ -55,11 +55,11 @@ class HealthSignalWindowActorSpec
           windowFreq = 1.second,
           initialWindowProcessingDelay = 10.milliseconds,
           actorSystem = system,
-          windowCheckInterval = 10.milliseconds).start(Some(probe.ref))
+          windowCheckInterval = 1.second).start(Some(probe.ref))
```
jeffboutotte, line 58: 1 second of artificial delay is way too long for these. If there's a weird timing thing with the shorter interval we should try to fix that rather than just increasing the time & making things slower artificially

zachary-rote: this isn't delay.  this is the interval on the `tick` in which a window is checked for expiry if no signals have been emitted.  Perhaps this calls for a better name.

jeffboutotte: Gotcha. What I really meant with the artificial delay was just increasing the test time artificially since we're waiting for those ticks as part of this test. Below our assertion is waiting for 5 tick messages. At a 1 second interval we're waiting 5 seconds for that (without the test actually really doing anything in that time we're waiting) where the lower interval would still test the same functionality, but our test isn't waiting around an extra 990ms per tick message it's expecting

jeffboutotte (approved):

modules/command-engine/core/src/test/scala/surge/internal/domain/SurgeMessagePipelineSpec.scala
```diff
@@ -29,7 +29,6 @@ import surge.metrics.Metrics
 
 import scala.concurrent.duration._
 
-@Ignore
```
jeffboutotte, line 32: Aah wait, actually just remembered another reason this was Ignored in an earlier PR. The logs end up flooded with this error message because something from this suite doesn't quite close cleanly:
```
2021-07-26 19:45:50,352 WARN  [kafka-producer-network-thread | producer-3] org.apache.kafka.clients.NetworkClient: [Producer clientId=producer-3] Connection to node 0 (localhost/127.0.0.1:6001) could not be established. Broker may not be available.
12315
```

Not sure if you have insights into what that might be?

PR #55, opened 2021-07-30
Timer refactor proposal

sebastianharko:
Currently the "timer" gets its own thread pool of 16 (!!) threads. I propose we get rid of it, and let the timer run on the implicit execution context.


jeffboutotte (approved):
LGTM - thanks!

PR #74, opened 2021-09-07
add getReplayProgress to ReplayCoordinatorApi in support of grpc repl…

zachary-rote:
add getReplayProgress to ReplayCoordinatorApi in support of grpc replay control impl

### DIFF
--- a/modules/common/src/main/scala/surge/internal/streams/ReplayCoordinator.scala
+++ b/modules/common/src/main/scala/surge/internal/streams/ReplayCoordinator.scala
@@ -69,11 +69,12 @@ class ReplayCoordinator(topicName: String, consumerGroup: String, registry: Acto
   override def receive: Receive = uninitialized()
 
   override def resumeConsumers(): Unit = {
-    self ! ResumeConsumers
+    context.self ! ResumeConsumers
   }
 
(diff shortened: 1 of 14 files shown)


modules/common/src/main/scala/surge/internal/streams/ReplayCoordinator.scala
```diff
@@ -71,6 +72,9 @@ class ReplayCoordinator(topicName: String, consumerGroup: String, registry: Acto
     self ! ResumeConsumers
   }
 
+  override def getReplayProgress: Future[ReplayProgress] =
+    self.ask(GetReplayProgress)(TimeoutConfig.ReplayCoordinatorActor.actorAskTimeout).mapTo[ReplayProgress]
```
jeffboutotte, line 76: Must have missed this in the other PR but the `self.ask` here is a bit odd. In general any communication to the actor should be through messages rather than direct method calls on it. If we want to separate out an API that's easy to consume from other places but back the logic by the actor we can follow a similar pattern to the `KafkaStreamManager` stuff.

zachary-rote: done

jeffboutotte (approved):

PR #77, opened 2021-09-08
Support async event handler

sebastianharko:


modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -211,10 +213,14 @@ class PersistentActor[S, M, R, E](
   private def freeToProcess(state: InternalActorState): Receive = {
     case pm: ProcessMessage[M] =>
       handle(state, pm)
-    case ae: ApplyEvent[E] => handle(state, ae)
-    case GetState(_)       => sender() ! StateResponse(state.stateOpt)
-    case ReceiveTimeout    => handlePassivate()
-    case Stop              => handleStop()
+    case ae: ApplyEvent[E] =>
```
sebastianharko, line 216: actually TBH I don't understand why we would ever need to handle this ApplyEvent message ? Since AggregateCommandModel can left fold over a number of events, why would we ever apply events one by one ...

sebastianharko: Cool, thanks for the explanation, I wasn't aware of the multi-DC aspect ;-)

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -263,29 +269,40 @@ class PersistentActor[S, M, R, E](
...
-      case Failure(e) =>
-        sender() ! ACKError(e)
+  private def callEventHandler(state: InternalActorState, evt: E): Future[HandleEventResult] = {
+    val result: Future[HandleEventResult] = {
+      metrics.eventHandlingTimer.time(businessLogic.model.applyAsync(surgeContext(), state.stateOpt, evt)).map { maybeS: Option[S] =>
+        HandleEventResult(result = maybeS)
+      }
     }
+    val timeOutFut =
+      akka.pattern.after(5.seconds, context.system.scheduler)(Future.failed(new Exception("Async event handler timed out after 5 seconds")))
```
jeffboutotte, line 279: Is it worth exposing a config setting for this timeout?

jeffboutotte (approved):
LGTM. One minor comment but I think this should be good to go

PR #80, opened 2021-09-09
gRPC multilanguage server

sebastianharko:


modules/multilanguage/src/main/protobuf/multilanguage-protocol.proto
```diff
@@ -0,0 +1,69 @@
...
+  string aggregateId = 1;
+}
+
+message GetStateReply {
+  string aggregateId = 1;
+  State state = 2;
+}
+
+service MultilanguageGatewayService {
+  rpc ForwardCommand (ForwardCommandRequest) returns (ForwardCommandReply) {}
```
jeffboutotte, line 62: Do we want to keep the naming here consistent with `sendCommand` or is there a subtle difference between send & forward?

modules/multilanguage/src/main/resources/logback.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
jeffboutotte, line 4: This works for now, but might be worth a follow up to allow the end user to configure log levels, log format, etc... for the grpc server as well.

modules/multilanguage/src/main/scala/com/ukg/surge/multilanguage/GenericAsyncAggregateCommandModel.scala
```diff
@@ -0,0 +1,56 @@
...
+    val reply: Future[ProcessCommandReply] = bridgeToBusinessApp.processCommand(processCommandRequest)
+
+    reply.flatMap { processCommandReply =>
+      {
+        if (processCommandReply.isSuccess) {
+          Future.successful {
+            processCommandReply.events.map(e => e: SurgeEvent)
+          }
+        } else {
+          Future.failed(new Exception(processCommandReply.rejectionMessage))
```
jeffboutotte, line 40: One potential place for some weirdness I could see would be where business logic to process commands throws an exception or returns a failed future explicitly rather than rejecting with a simple message. Maybe the SDK can handle the wrapping of that somehow. Just thoughts for the future - this is good enough for now to at least prove out that it works

modules/multilanguage/src/main/scala/com/ukg/surge/multilanguage/MultilanguageGatewayServiceImpl.scala
```diff
@@ -0,0 +1,79 @@
...
+    in.command match {
+      case Some(cmd: protobuf.Command) =>
+        logger.info(s"Received command for aggregate with id ${cmd.aggregateId}, payload has size ${cmd.payload.size()} (bytes)!")
+        val aggIdStr = cmd.aggregateId
+        val aggIdUUID: UUID = UUID.fromString(aggIdStr)
+        val surgeCmd: SurgeCmd = cmd
+        logger.info(s"Forwarding command to aggregate with id ${cmd.aggregateId}!")
+        surgeEngine.aggregateFor(aggIdUUID).sendCommand(surgeCmd).map {
+          case CommandSuccess(aggregateState: Option[SurgeState]) =>
+            val payloadSize: Any = aggregateState.map(_.payload.length).getOrElse(0)
```
jeffboutotte, line 54: Isn't this an `Int` or `Long` rather than just `Any`?

modules/multilanguage/src/main/scala/com/ukg/surge/multilanguage/MultilanguageGatewayServiceImpl.scala
```diff
@@ -0,0 +1,79 @@
...
+    logger.info("Started engine!")
+    engine
+  }
+
+  override def forwardCommand(in: ForwardCommandRequest): Future[ForwardCommandReply] = {
+    in.command match {
+      case Some(cmd: protobuf.Command) =>
+        logger.info(s"Received command for aggregate with id ${cmd.aggregateId}, payload has size ${cmd.payload.size()} (bytes)!")
+        val aggIdStr = cmd.aggregateId
+        val aggIdUUID: UUID = UUID.fromString(aggIdStr)
```
jeffboutotte, line 49: Might be worth typing the engine & business logic on `String` for AggIdType rather than forcing it to always be a `UUID` for a little more flexibility. That'd simplify these bits a little as well

jeffboutotte (approved):
A few thoughts for later we can revisit as we continue to prove out that this pattern works

modules/multilanguage/src/main/scala/com/ukg/surge/multilanguage/MultilanguageGatewayServiceImpl.scala
sebastianharko: you are right not sure it shows up as Any, i'll investigate

modules/multilanguage/src/main/protobuf/multilanguage-protocol.proto
sebastianharko: @jeffboutotte The mental I have in my mind is that the gRPC server is "forwarding" the command to Surge and Surge is "sending" the command to the right actor on the right node ...Let me know what you think and I don't mind renaming to sendCommand because it's not a big change, totally up to you .

PR #87, opened 2021-09-13
Hotfix/offset null fix

zachary-rote:


modules/common/src/main/scala/surge/streams/replay/EventReplayStrategy.scala
```diff
@@ -98,6 +98,10 @@ trait ReplayControl {
       consumerGroup: String,
       partitions: Iterable[Int],
       replayLifecycleCallbacks: ReplayLifecycleCallbacks = new NoopReplayLifecycleCallbacks()): Future[Done]
+
+  private def sum(offsets: Map[TopicPartition, OffsetAndMetadata]): Long = {
+    offsets.values.map[Long](o => Option(o).map[Long](offsetPlusMeta => offsetPlusMeta.offset()).getOrElse(0L)).sum
```
jeffboutotte, line 103: Wait `OffsetAndMetadata` were coming back as `null` in here? That's strange. I think you can use a `flatMap` here as well. Something like:
```
offsets.values.flatMap(o => Option(o).map(_.offset())).sum
```

jeffboutotte (approved):

PR #92, opened 2021-09-16
Docs: fix the BankAccountEngine

EldarKZ16:
If the surge engine is not started, it throws timeout exception


jeffboutotte (approved):
Cool - makes sense to me. Thanks!

PR #93, opened 2021-09-16
report progress as 100.0% when replaying an empty stream

zachary-rote:


modules/common/src/test/scala/surge/internal/streams/StreamManagerSpec.scala
```diff
@@ -218,6 +218,33 @@ class StreamManagerSpec
...
+        val kafkaForeverReplayStrategy =
+          new KafkaForeverReplayStrategyImpl(defaultConfig, system, settings, () => Future.successful(true), () => completeProbe.ref ! ReplayComplete)
+        val consumer =
+          testStreamManager(topic, kafkaBrokers = embeddedBroker, groupId = "replay-test", sendToTestProbe(probe), kafkaForeverReplayStrategy, settings)
+
+        consumer.start()
+
+        // fix: if we attempt to replay prior to a stream manager being registered for the topic replay will fail.
+        //  we need a to ensure the streamManager is started prior to calling replay.
+        Thread.sleep(5.seconds.toMillis)
```
jeffboutotte, line 239: Good for now but we'll want to circle back and make this more explicit - we could send it an "Are you ready" type message and assert we get a response back rather than the sleep here.

jeffboutotte (approved):

PR #98, opened 2021-09-21
Add scala sample

EldarKZ16:


jeffboutotte (commented):
There may be some duplication between this and the [paradox docs](https://github.com/UltimateSoftware/surge/tree/main/modules/surge-docs/src/test/scala/docs/command). Ideally for docs we should be pulling examples from that surge-docs module since we can actually demonstrate & test functionality there as part of the build. At some point it may also be worth having a couple surge-examples projects for real runnable references. Once we hit a stable 1.0 we can create a surge-examples repo for those, but for now if that's what this wants to be maybe we want this (and the java example) to just be submodules in the top level build.sbt that get built alongside everything else to ensure we're keeping everything up to date.

jeffboutotte (approved):
LGTM - thanks!

PR #99, opened 2021-09-21
Add health check grpc command

EldarKZ16:


modules/multilanguage-protocol/src/main/protobuf/multilanguage-protocol.proto
```diff
@@ -65,12 +66,21 @@ message GetStateReply {
...
+  string status = 1;
+}
+
 service MultilanguageGatewayService {
   rpc ForwardCommand (ForwardCommandRequest) returns (ForwardCommandReply) {}
   rpc GetState (GetStateRequest) returns (GetStateReply) {}
 }
 
 service BusinessLogicService {
+  rpc HealthCheckCommand (HealthCheckRequest) returns (HealthCheckReply) {}
```
sebastianharko, line 83: @EldarKZ16 I think you can call this just HealthCheck instead of HealthCheckCommand

jeffboutotte: I think we'd want to health check the sidecar so this might want to be on the `MultilanguageGatewayService`?

sebastianharko: @jeffboutotte we'll eventually add it to both .... the Surge sidecar ("multilanguage server") is going to ping the business app ... and the business app is going to ping the Surge sidecar ....

modules/multilanguage/src/main/scala/com/ukg/surge/multilanguage/MultilanguageGatewayServiceImpl.scala
```diff
@@ -86,4 +86,7 @@ class MultilanguageGatewayServiceImpl(aggregateName: String, eventsTopicName: St
 
   }
 
+  override def healthCheck(in: HealthCheckRequest): Future[HealthCheckReply] =
+    bridgeToBusinessApp.healthCheck(in)
```
jeffboutotte, line 90: I think here you could use the `surgeEngine.healthCheck` method to check the healthiness of the actual engine. It'll include a bunch of details that you can either include (and expand the protobuf model for the `HealthCheckReply`) or filter out to distill an `UP` or `DOWN` status from the top level of the health check tree.

EldarKZ16: Added

sebastianharko: @jeffboutotte I agree ...So the healthcheck for this has two conditions..1. is the underlying surge engine healthy (and here we use the surgeEngine.healthCheck) ....and 2. can we connect to the business app via gRPC (and here we use the new gRPC call)

garyKeorkunian:
Is this one ready to go?

sebastianharko:
> Is this one ready to go?

Not yet @garyKeorkunian

modules/multilanguage/src/main/scala/com/ukg/surge/multilanguage/MultilanguageGatewayServer.scala
```diff
@@ -7,39 +7,59 @@ import akka.event.{ Logging, LoggingAdapter }
...
+  }
 
-    logger.info(
-      s"Binding multilanguage gateway server on $host:$port. Aggregate name: $aggregateName. Events topic: $eventsTopicName. State topic: $stateTopicName")
+  def doHealthCheck(): Future[HealthCheckReply] = {
+    for {
+      surgeHealth <- surgeEngine.healthCheck
+      businessAppHealth <- bridgeToBusinessApp.healthCheck(HealthCheckRequest()) if surgeHealth.isHealthy.getOrElse(false)
+      multilanguageServerHealth = businessAppHealth.copy(serviceName = "multilanguage-server") if businessAppHealth.status == Status.UP
+    } yield multilanguageServerHealth
```
jeffboutotte, line 50: Hmm what happens on this if `surgeHealth` comes back unhealthy here? With the if's in here I'm wondering if we yield a `Future.successful(HealthCheckReply(status = DOWN))` or a `Future.failed(not sure what it'd fail with?)`

modules/multilanguage/src/main/scala/com/ukg/surge/multilanguage/MultilanguageSidecarMain.scala
```diff
@@ -4,11 +4,38 @@ package com.ukg.surge.multilanguage
...
+  implicit val ec: ExecutionContext = system.dispatcher
+
   val logger = Logging(system, classOf[MultilanguageSidecarMain])
-  val binding = new MultilanguageGatewayServer(system).run()
+  val multilanguageServer = new MultilanguageGatewayServer(system)
+
+  val route = path("healthz") {
+    get {
+      onComplete(multilanguageServer.doHealthCheck()) {
+        case Success(_)  => complete(StatusCodes.OK)
```
jeffboutotte, line 27: With the response here I think we can look at whether or not the status is up or down to determine status code we reply with in addition to the success/failure of the future

EldarKZ16: Actually, I did this because of kubernetes livenessProbe. 
It will just check the status code of the response. 
And I've updated the doHealthCheck.

jeffboutotte (approved):
LGTM - thanks!

PR #106, opened 2021-09-23
fix flaky producer fencing spec

zachary-rote:


modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -171,6 +171,8 @@ class KafkaProducerActorImpl(
     case InitTransactions       => initializeTransactions()
     case InitTransactionSuccess => initTransactionsSuccess(lastProgressUpdate)
     case FlushMessages          => log.trace("KafkaPublisherActor ignoring FlushMessages message from the uninitialized state")
+    case failed: EventsFailedToPublish =>
```
jeffboutotte, line 174: I think this is still a subtle bug/potential race condition here. In practice if we hit this case the `PersistentActor` would be stuck waiting for a reply and eventually time out even though we know here there's a failure and could explicitly reply back to it. Given that it's still a failure to publish messages regardless of what state inside the producer actor handles the failure I think we can just have the same behavior everywhere of notifying anyone waiting with a `PublishFailure` message.

zachary-rote: done

modules/command-engine/core/src/test/scala/surge/internal/kafka/KafkaProducerActorImplSpec.scala
```diff
@@ -509,7 +509,9 @@ class KafkaProducerActorImplSpec
       probe.watch(fencedOnCommit)
       probe.send(fencedOnCommit, KafkaProducerActorImpl.Publish(testAggs1, testEvents1))
       probe.send(fencedOnCommit, KafkaProducerActorImpl.FlushMessages)
-      probe.expectMsgType[PublishFailure]
+      val result = probe.fishForMessage(10.seconds) { msg => msg.isInstanceOf[PublishResult] }
```
jeffboutotte, line 512: I think this can just be the regular `probe.expectMsgType` as it was before since it should always be a `PublishFailure`

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
jeffboutotte: Cool this LGTM. We may want this same behavior in the `waitingForKTableIndexing` state as well in case it happens to transition to that state before getting the `EventsFailedToPublish` message. I think that'd be an even smaller window to get into that scenario, but doesn't hurt to make sure we're covered there

modules/command-engine/core/src/test/scala/surge/internal/kafka/KafkaProducerActorImplSpec.scala
zachary-rote: done

jeffboutotte (approved):
LGTM

PR #109, opened 2021-09-27
MDC context propagation in futures

satyender071:
Added the MDC context propagation in scala futures.
Added the test cases also.

### DIFF
--- a/modules/common/src/main/scala/surge/internal/utils/MdcExecutionContext.scala
+++ b/modules/common/src/main/scala/surge/internal/utils/MdcExecutionContext.scala
@@ -1,0 +1,11 @@
+package surge.internal.utils
+
+import java.util.concurrent.Executors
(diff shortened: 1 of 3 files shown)


modules/common/src/main/scala/surge/internal/utils/MdcExecutionContext.scala
```diff
@@ -0,0 +1,11 @@
+package surge.internal.utils
+
+import java.util.concurrent.Executors
+import scala.concurrent.ExecutionContext
+
+object MdcExecutionContext {
```
sebastianharko, line 6: I don't think you need this file ;-)

satyender071: Changed it with scala global execution context

garyKeorkunian (approved):

PR #111, opened 2021-09-28
fix flaky surgemessagepipelinespec

zachary-rote:


garyKeorkunian (approved):

jeffboutotte:
I think something in this spec isn't shutting down properly - the logs end up flooded with messages that look like:
```
2021-09-30 12:48:38,183 WARN  [kafka-producer-network-thread | producer-2] org.apache.kafka.clients.NetworkClient: [Producer clientId=producer-2] Connection to node 0 (localhost/127.0.0.1:6001) could not be established. Broker may not be available.
5767
```

PR #116, opened 2021-09-30
Add Concurrency setting to GH Workflows

garyKeorkunian:
This change will cause previous in-progress builds to cancel whenever a push is made to the PR branch


jeffboutotte (approved):
Cool! Good find

PR #128, opened 2021-10-26
Akka Cluster Sharding

EldarKZ16:


modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
```diff
@@ -27,25 +34,78 @@ private[surge] final class SurgePartitionRouterImpl(
...
+    shardRouter
+  }
+
+  private def createAkkaCluster(): ActorRef = {
+    val groupId = businessLogic.kafka.streamsApplicationId
+    val consumerSettings = ConsumerSettings(system, new StringDeserializer, new ByteArrayDeserializer)
+      .withBootstrapServers(config.getString("kafka.brokers"))
+      .withGroupId(groupId)
+      .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
+      .withStopTimeout(0.seconds)
```
jeffboutotte, line 77: This is one of the interesting pieces I'm curious on. The Alpakka Kafka docs have the copartitioning based on the input topic so you can leverage the same consumer settings as your consumer, but with Surge we're copartitioning based on the Surge Kafka Streams consumer (totally outside of Akka). I'm wondering if under the hood the Akka KafkaClusterSharding is doing something special inside the consumer group we should be aware of or if it just leverages the groupId (and any connection related settings) to grab what it needs?

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -58,6 +59,59 @@ object KafkaProducerActor {
...
 
+  def create(
+      actorSystem: ActorSystem,
+      metrics: Metrics,
+      businessLogic: SurgeModel[_, _, _, _],
+      kStreams: AggregateStateStoreKafkaStreams[_],
+      partitionTracker: KafkaConsumerPartitionAssignmentTracker,
+      signalBus: HealthSignalBusTrait,
+      config: Config,
+      kafkaProducerOverride: Option[KafkaProducerTrait[String, Array[Byte]]] = None): Int => KafkaProducerActor = partitionNumber => {
```
jeffboutotte, line 70: `create` is a little confusing for a name here since this returns a function that creates producer actors based on a partition number rather than a producer actor directly. Maybe something like `createFromPartitionFunction`? I'm open to better suggestions as well :)

modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
jeffboutotte: Also I think we'll need to plumb authentication related settings to this to support auth for Kafka

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -147,6 +150,13 @@ class PersistentActor[S, M, R, E](
   import PersistentActor._
   import context.dispatcher
 
+  //FIXME: temporary fix to support switch between akka and existing shard allocation strategy
+  def aggregateId: String = aggregateIdOpt.getOrElse(self.path.name)
```
jeffboutotte, line 154: Actually this isn't a bad assumption - in almost any real-world case `aggregateIdOpt` should be `Some(aggId)` and the actor itself is named by the aggregateId

modules/command-engine/core/src/test/scala/surge/internal/domain/SurgeMessagePipelineSpec.scala
```diff
@@ -295,14 +296,17 @@ class SurgeMessagePipelineSpec
 
       override def actorSystem: ActorSystem = system
 
+      private val isAkkaClusterEnabled = Try(config.getBoolean("surge.akka.cluster.enabled")).getOrElse(false)
```
jeffboutotte, line 299: The `Try` here seems overprotective - we should define the setting in our reference.conf so it should always have a value

modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
EldarKZ16: Yeah, it just leverages connection-related settings. And basically, we use this kafka consumer setting to extract the number of partitions for the state topic.

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -60,6 +61,62 @@ object KafkaProducerActor {
...
+        metrics = metrics,
+        businessLogic,
+        lagChecker = new KTableLagCheckerImpl(kStreams.applicationId, adminClient),
+        partitionTracker = partitionTracker,
+        signalBus = signalBus,
+        config = config,
+        kafkaProducerOverride = kafkaProducerOverride)).withDispatcher(dispatcherName)
+
+    new KafkaProducerActor(
+      ActorLifecycleManagerActor.manage(
```
jeffboutotte, line 91: Can you use named args for all of these to make it more obvious what maps to what options?

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaClusterShardingRebalanceListener.scala
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  private def handle(state: ActorState, partitionAssignments: PartitionAssignments): Unit = {
+    val newStateWithPartitionAssignments = state.updatePartitionAssignments(partitionAssignments)
+
+    val updatedTopicPartitions = newStateWithPartitionAssignments.partitionAssignments.partitionAssignments.collect {
+      case (hostPort, topicPartitions) if isHostPortThisNode(hostPort) => topicPartitions
+    }.flatten
+
+    val updates = shardAllocationClient.updateShardLocations(updatedTopicPartitions.map { tp =>
```
jeffboutotte, line 148: Should `updateShardLocations` here be updating the locations of all of the shards including those assigned to other nodes in order for the cluster to be aware of which nodes are assigned which shards across the board? Right now it's filtering out and only providing updates on the partitions that are assigned to this node

jeffboutotte (commented):
One last round of comments - once they're addressed I think we're good to merge this in and start doing some more testing of it

project/Dependencies.scala
```diff
@@ -8,7 +8,16 @@ object Dependencies extends AutoPlugin {
     object Akka {
       val version = "2.6.15"
       val akkaHttpVersion = "10.2.6"
+      val alpakkaVersion = "2.1.1"
+      val managementVersion = "1.1.1"
 
+      val kafkaStream = "com.typesafe.akka" %% "akka-stream-kafka" % alpakkaVersion
```
jeffboutotte, line 14: This already exists as Alpakka.kafka but I think Akka.kafkaStream makes more sense so we can clean up the old one

modules/common/src/main/scala/surge/internal/tracing/SurgePersistentEntity.scala
```diff
@@ -1,6 +1,6 @@
 // Copyright © 2017-2021 UKG Inc. <https://www.ukg.com>
 
-package surge.internal.persistence
+package surge.internal.tracing
```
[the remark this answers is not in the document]
EldarKZ16: The TracedMessage needed to extend RoutableMessage, and it wasn't available from the persistence package.

modules/command-engine/core/src/main/scala/surge/internal/domain/SurgeMessagePipeline.scala
```diff
@@ -122,6 +126,34 @@ private[surge] abstract class SurgeMessagePipeline[S, M, +R, E](
 
   override def shutdown(): Future[Ack] = stop()
 
+  private def startClusterManagementAndRebalanceListener(): Future[Unit] = {
+    if (isAkkaClusterEnabled) {
+      Cluster.get(system)
+      ClusterBootstrap(system).start()
+      for {
+        _ <- AkkaManagement(system).start()
```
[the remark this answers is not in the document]
EldarKZ16: I've added a new class called KafkaStreamsServiceDiscovery. It identifies other nodes from Kafka consumer metadata. 
But after identifying the IP address, it probes HTTP request to another node to see if a cluster has already been formed.  
So it works as it is said in the [docs](https://doc.akka.io/docs/akka-management/current/bootstrap/index.html).

modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
```diff
@@ -2,48 +2,108 @@
...
-  private val routerActorName = s"${businessLogic.aggregateName}RouterActor"
+    val routerActorName = s"${businessLogic.aggregateName}RouterActor"
+    val shardRouter = system.actorOf(shardRouterProps, name = routerActorName)
+    shardRouter
+  }
+
+  private def createAkkaCluster(): ActorRef = {
+    val groupId = businessLogic.kafka.streamsApplicationId
+    val consumerSettings = ConsumerSettings(system, new StringDeserializer, new ByteArrayDeserializer)
+      .withBootstrapServers(config.getString("kafka.brokers"))
```
EldarKZ16, line 73: We can't leverage the KafkaSecurityConfiguration in this place. 
I've analyzed the [codebase](**url**) and it seems like we can't configure it manually. 
Possibly, we can pass security configs as described in this [docs](https://doc.akka.io/docs/alpakka-kafka/current/production.html)

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaClusterShardingRebalanceListener.scala
EldarKZ16: Each node will update its own shard locations and co-locate it with assigned kafka partitions. 
Same thing happens in [akka-kafka-cluster-sharding library](https://github.com/akka/alpakka-kafka/blob/8762635be58b83a4cab9194228343e34b9aa7c60/cluster-sharding/src/main/scala/akka/kafka/cluster/sharding/KafkaClusterSharding.scala#L309) which uses alpakka kafka consumer

modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
jeffboutotte: You should be able to pass in just a raw properties map via `.withProperties` - something like:
```
val securityRelatedProps = new Properties()
val securityHelper = new KafkaSecurityConfigurationImpl(config)
securityHelper.configureSecurityProperties(securityRelatedProps)

consumerSettings
  .withProperties(securityRelatedProps.asScala.toMap)
```

EldarKZ16: Missed that, thank you! I've updated the kafka consumer settings

modules/common/src/main/scala/surge/internal/tracing/SurgePersistentEntity.scala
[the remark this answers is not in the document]
jeffboutotte: Aah I see, it moved from command-engine-core to common as well. Thanks!

modules/command-engine/core/src/main/scala/surge/internal/domain/SurgeMessagePipeline.scala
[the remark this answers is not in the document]
jeffboutotte: This is ok for now I think (especially given this is experimental and needs to be enabled explicitly based on that setting).

jeffboutotte (dismissed):

jeffboutotte (approved):

PR #131, opened 2021-10-28
Apply events in bulk

ckmcd:


modules/command-engine/javadsl/src/main/scala/surge/javadsl/common/AggregateRefBaseTrait.scala
```diff
@@ -24,8 +24,8 @@ trait AggregateRefBaseTrait[AggId, Agg, Cmd, Event] extends AggregateRefTrait[Ag
     FutureConverters.toJava(queryState.map(_.asJava))
   }
 
-  def applyEvent(event: Event): CompletionStage[ApplyEventResult[Agg]] = {
-    val envelope = PersistentActor.ApplyEvent[Event](aggregateId.toString, event)
+  def applyEvents(events: Seq[Event]): CompletionStage[ApplyEventResult[Agg]] = {
```
jeffboutotte, line 27: I think we want this to remain backwards compatible from here with both the old `applyEvent` and new `applyEvents` method

modules/command-engine/scaladsl/src/main/scala/surge/scaladsl/command/AggregateRef.scala
```diff
@@ -17,7 +17,7 @@ trait AggregateRef[Agg, Cmd, Event] {
 
   def sendCommand(command: Cmd): Future[CommandResult[Agg]]
 
-  def applyEvent(event: Event): Future[ApplyEventResult[Agg]]
+  def applyEvents(event: Seq[Event]): Future[ApplyEventResult[Agg]]
```
jeffboutotte, line 20: Same here - keep the old `applyEvent` around for backwards compatibility

modules/command-engine/javadsl/src/main/scala/surge/javadsl/common/AggregateRefBaseTrait.scala
jeffboutotte: Aah never mind, I see now :)

jeffboutotte (dismissed):
Just that one change to the scaladsl to keep things consistent between the javadsl and scaladsl but this looks good otherwise

jeffboutotte (dismissed):

jeffboutotte (approved):

PR #134, opened 2021-11-03
Expose Java future from surge engine stop in javadsl

satyender071:
Added a `surgeEngine.stop()` to return `CompletionStage` in javadsl


modules/common/src/main/scala/surge/core/Controllable.scala
```diff
@@ -17,7 +17,7 @@ trait ControllableRemover {
 trait Controllable {
```
jeffboutotte, line 17: Hmm I'm not sure I'm crazy about `stopInternal` now that I'm seeing it in this change but maybe `Controllable` should be internal and we can expose an explicit `start`/`stop` from the javadsl and scaladsl.

satyender071: I think that would need  a discussion cause we have to change it for every method in the `contollable` trait. we van discuss in the stand up call

sebastianharko (approved):

sebastianharko (approved):

jeffboutotte (approved):

PR #135, opened 2021-11-08
Properly Propagate MDC and Spans Across Nodes & Execution Contexts

jeffboutotte:
Propagates spans and MDC across various execution contexts so that the business logic can properly create child spans and so that traces are able to be continued when executing on different threads.

### DIFF
--- a/modules/common/src/main/scala/surge/internal/utils/DiagnosticContextFuturePropagation.scala
+++ b/modules/common/src/main/scala/surge/internal/utils/DiagnosticContextFuturePropagation.scala
@@ -2,10 +2,11 @@
 
 package surge.internal.utils
 
-import io.opentelemetry.api.trace.Span
+import io.opentelemetry.api.trace.{ Span, Tracer }
+import io.opentelemetry.context.Context
 import org.slf4j.MDC
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ ExecutionContext, ExecutionContextExecutor }
 
 /**
  * Execution context proxy for propagating SLF4J diagnostic context and the current span from caller thread to execution thread.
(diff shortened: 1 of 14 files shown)


scalastyle-config.xml
```diff
@@ -94,7 +94,7 @@
  </check>
  <check level="warning" class="org.scalastyle.scalariform.NumberOfMethodsInTypeChecker" enabled="true">
   <parameters>
-   <parameter name="maxMethods"><![CDATA[30]]></parameter>
+   <parameter name="maxMethods"><![CDATA[40]]></parameter>
```
jeffboutotte, line 97: Probably want to revert this...`PersistentActor` has more than 30 methods (which is a lot) but it's distracting to edit in the IDE. I think the fact it's setting off this scalastyle warning is a red flag and we should circle back and clean it up a bit if possible.

modules/common/src/main/scala/surge/internal/utils/DiagnosticContextFuturePropagation.scala
```diff
@@ -14,15 +15,20 @@ class DiagnosticContextFuturePropagation(executionContext: ExecutionContext) ext
   override def execute(runnable: Runnable): Unit = {
     val callerMdc = Option(MDC.getCopyOfContextMap) // get a copy of the MDC context data
     val callerSpan = Span.current()
+    println(s"JEFF --- caller MDC = $callerMdc")
+    println(s"JEFF --- caller span = $callerSpan")
```
jeffboutotte, line 19: A bunch of these printlns will need to be cleaned up. I was using for debug but they ultimately don't belong in the code

jeffboutotte:
This is a bit stale, will pull relevant changes into a new branch.

PR #137, opened 2021-11-16
experimental annotations

zachary-rote:


sebastianharko:
@zachary-rote I suggest we also add something like "ApiMayChange" to indicate that a feature is not experimental but its API may change. This is for cases when what's under the hood is solid but the user facing API is not quite ready..

sebastianharko (approved):

jeffboutotte (approved):

PR #139, opened 2021-11-23
Added a SurgeContext aware message processing model

jeffboutotte:
This model should simplify the amount of layering that we have in Surge where the various processing models inherit from one another and inevitably end up wired incorrectly due to behaviors changing unexpectedly through the various layers.


modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -406,7 +406,7 @@ class KafkaProducerActorImpl(
 
   private def publishSingleRecord(senders: Seq[ActorRef], record: ProducerRecord[String, Array[Byte]]): Future[InternalMessage] = {
     kafkaPublisherTimer.timeFuture {
-      kafkaPublisher
+      nonTransactionalStatePublisher
```
jeffboutotte, line 409: This could get pulled out into its own change - it fixes a bug with the experimental setting for disabling transactions

build.sbt
```diff
@@ -62,6 +62,7 @@ lazy val `surge-engine-command-core` = (project in file("modules/command-engine/
   .settings(libraryDependencies ++= Seq(
     Akka.actor,
     Akka.remote,
+    Akka.persistenceTyped,
```
jeffboutotte, line 65: Remove this since it's unused

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
```diff
@@ -57,20 +57,22 @@ trait KTablePersistenceSupport[Agg, Event] {
 
   protected def doPublish(
       state: ActorState,
+      context: SurgeContextImpl[Agg, Event], // FIXME add serialized events/state to the context???
```
jeffboutotte, line 60: Maybe address this FIXME

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -234,69 +223,85 @@ class PersistentActor[S, M, R, E](
...
     context.setReceiveTimeout(Duration.Inf)
     context.become(persistingEvents(state))
 
     processMessage(state, msg)
-      .flatMap {
-        case Left(r) => Future.successful(ACKRejection(r))
-        case Right(handled) =>
+      .flatMap { result =>
+        if (result.isRejected) {
+          // FIXME Temporary no-op publish to trigger side effects later
```
jeffboutotte, line 233: This is a bit awkward and warrants some cleanup

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
```diff
@@ -47,8 +47,11 @@ object PersistentActor {
       @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "aggregateType", visible = true) aggregateState: Option[S])
       extends JacksonSerializable
 
-  sealed trait ACK extends ActorMessage with JacksonSerializable
+  sealed trait ACK
+      extends ActorMessage
+      with JacksonSerializable
 
+      // FIXME This is an explicit response type now
```
jeffboutotte, line 54: Need some renaming as this is no longer necessarily `aggregateState` but a generic `response`

modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/CommandModels.scala
```diff
@@ -2,109 +2,77 @@
...
   def handleEvent(aggregate: Optional[Agg], event: Evt): Optional[Agg]
 
-  final def toCore: CommandHandler[Agg, Cmd, Nothing, Evt] =
-    new CommandHandler[Agg, Cmd, Nothing, Evt] {
-      override def processCommand(ctx: persistence.Context, state: Option[Agg], cmd: Cmd): Future[CommandResult] =
-        Future.fromTry(Try(Right(AggregateCommandModel.this.processCommand(state.asJava, cmd).asScala.toSeq)))
-      override def apply(ctx: persistence.Context, state: Option[Agg], event: Evt): Option[Agg] = handleEvent(state.asJava, event).asScala
+  final def toCore: SurgeProcessingModel[Agg, Cmd, Evt] =
+    new SurgeProcessingModel[Agg, Cmd, Evt] {
+      override def handle(ctx: SurgeContext[Agg, Evt], state: Option[Agg], msg: Cmd)(implicit ec: ExecutionContext): Future[SurgeContext[Agg, Evt]] = {
```
jeffboutotte, line 23: This is how the new context based processing model works - from the context you can `persistEvents` `updateState` and `reply` however you'd like. This should allow for a lot of flexibility down the road without us needing to plumb additional types and processing models all the way through to the core.

modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/CommandModels.scala
```diff
@@ -2,109 +2,77 @@
...
+        Future.fromTry(Try(AggregateCommandModel.this.processCommand(state.asJava, msg).asScala.toSeq)).map { events =>
+          val newState = events.foldLeft(state.asJava)((s: Optional[Agg], e: Evt) => handleEvent(s, e))
+          ctx.persistEvents(events).updateState(newState.asScala).reply(state => state)
+        }
+      }
+
+      override def applyAsync(ctx: SurgeContext[Agg, Evt], state: Option[Agg], event: Evt): Future[SurgeContext[Agg, Evt]] = {
```
jeffboutotte, line 30: This maybe wants to be revisited since the state update now happens through the `handle` method and we don't call `applyAsync` except for when applying events to the state directly (outside of the normal command flow). Ideally we can find some decent balance between a CQRS model (command handler and event handler) and generic processing model (handle this message and tell the context what you'd like to do)

modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/CommandModels.scala
```diff
@@ -2,109 +2,77 @@
...
-        FutureConverters.toScala(RejectableAggregateCommandModel.this.processCommand(Context(ctx), state.asJava, cmd))
-      override def apply(ctx: persistence.Context, state: Option[Agg], event: Evt): Option[Agg] = handleEvent(Context(ctx), state.asJava, event).asScala
+      override def applyAsync(ctx: SurgeContext[Agg, Evt], state: Option[Agg], event: Evt): Future[SurgeContext[Agg, Evt]] = {
+        val newState = handleEvent(state.asJava, event)
+        Future.successful(ctx.updateState(newState.asScala))
+      }
     }
 }
+
+// TODO Add explicit reply type model to prove out the abstraction
```
jeffboutotte, line 78: Need to do this TODO still

jeffboutotte: Also need to add back a model where the domain can explicitly reject commands somehow

modules/command-engine/scaladsl/src/main/scala/surge/scaladsl/command/CommandModels.scala
```diff
@@ -2,97 +2,73 @@
...
+  final def toCore: SurgeProcessingModel[Agg, Cmd, Evt] = {
+    new SurgeProcessingModel[Agg, Cmd, Evt] {
+      override def handle(ctx: SurgeContext[Agg, Evt], state: Option[Agg], msg: Cmd)(implicit ec: ExecutionContext): Future[SurgeContext[Agg, Evt]] = {
+        Future.fromTry(processCommand(state, msg)).map { events =>
+          val newState = events.foldLeft(state)((s: Option[Agg], e: Evt) => handleEvent(s, e))
+          ctx.persistEvents(events).updateState(newState).reply(state => state)
+        }
+      }
+
+      // FIXME applyAsync everywhere needs a Future.fromTry to handle exceptions thrown by the domain
```
jeffboutotte, line 25: Need to double check this. This was being (sorta magically) handled by one of the layers before but must be handled at this layer everywhere now

project/Dependencies.scala
```diff
@@ -20,6 +20,7 @@ object Dependencies extends AutoPlugin {
       val testKit = "com.typesafe.akka" %% "akka-testkit" % version % Test
       val akkaStreamTestKit = "com.typesafe.akka" %% "akka-stream-testkit" % version % Test
       val jacksonSerialization = "com.typesafe.akka" %% "akka-serialization-jackson" % version
+      val persistenceTyped = "com.typesafe.akka" %% "akka-persistence-typed" % version
```
jeffboutotte, line 23: Remove this since it's unused

modules/surge-docs/src/test/scala/docs/command/BankAccountEngine.scala
```diff
@@ -8,7 +8,7 @@ import surge.scaladsl.command.SurgeCommand
 
 // #bank_account_engine_class
 object BankAccountEngine {
-  lazy val surgeEngine: SurgeCommand[UUID, BankAccount, BankAccountCommand, Nothing, BankAccountEvent] = {
+  lazy val surgeEngine: SurgeCommand[UUID, BankAccount, BankAccountCommand, BankAccountEvent] = {
```
jeffboutotte, line 11: TODO This breaks backwards compatibility but was a bit awkward anyways, take a look at the impact of this change and possibly revert it for now

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
jeffboutotte: I'd like to revisit the `PersistentActor` as a whole eventually anyways so it's likely any optimization here now would end up deleted as part of that work. Additionally, we're not currently exposing a model where commands can be explicitly rejected, the current pattern (though maybe a little wacky) is to throw an exception from the command handler and treat it as an error.

modules/command-engine/core/src/main/scala/surge/internal/persistence/PersistentActor.scala
jeffboutotte: Will revisit this as part of the PersistentActor cleanup later as well...

modules/surge-docs/src/test/scala/docs/command/BankAccountEngine.scala
jeffboutotte: The way things are layered today this breaks compatibility in the open source layers but would not impact users internal to UKG since this particular API is hidden behind one flavored with UKG specific types. If we had these artifacts publicly published we'd want to be more careful of preserving compatibility here, but I think it's fine to break this API here especially while we know of and can limit the exact impact.

garyKeorkunian (dismissed):

garyKeorkunian (approved):

PR #140, opened 2021-11-24
PersistentActor.ACKError should be undergo serialization verification

EldarKZ16:
https://github.com/UltimateSoftware/surge-ukg/issues/103


EldarKZ16:
I haven't found ways how we can deserialize/serialize Throwables using jackson json. It keeps showing warning that illegal reflective access has been made to Throwable.detailMessage. 
So I thought we can return only the error message as a string, because we are logging all Throwables before sending it to the sender actor.

modules/common/src/main/scala/surge/exceptions/SurgeException.scala
```diff
@@ -0,0 +1,9 @@
+package surge.exceptions
+
+sealed trait SurgeException {
+  def message: String
+
+  def toJavaException: Exception = new Exception(message)
+}
+
+case class SurgeCustomException(message: String) extends SurgeException
```
jeffboutotte, line 9: Any way we can actually wrap the exception here instead of just the message string? Having the actual exception type + stack trace is actually really useful for apps who may reject commands with explicit exceptions that they can look for by explicit type on the asking end - for example in our example bank account we fail with an `InsufficientFundsException` if we try to withdraw too much from an account: https://github.com/UltimateSoftware/surge-samples/blob/main/examples/scala/src/main/scala/com/example/model/BankAccountCommandModel.scala#L32

modules/command-engine/core/src/test/scala/surge/internal/persistence/PersistentActorSpec.scala
```diff
@@ -610,5 +610,26 @@ class PersistentActorSpec
       doSerde(PersistentActor.ApplyEvent[String](UUID.randomUUID().toString, "test2"))
       doSerde(eventEnvelope(CountIncremented(UUID.randomUUID().toString, 1, 1)))
     }
+
+    "Serialize/Deserialize ACK types from Akka" in {
+      import akka.serialization.SerializationExtension
+
+      def doSerde[A](envelope: ACK): Unit = {
```
jeffboutotte, line 617: This `doSerde` function is duplicated across these few tests for serializing/deserializing various types (request & response) - I think we can extract it to a `def doAkkaSerde(message: AnyRef): Assertion` to eliminate a bit of the duplication

modules/common/src/main/scala/surge/exceptions/SurgeException.scala
EldarKZ16: If we wrap the exception, it works but the jackson shows these kinds of warnings
`WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.fasterxml.jackson.databind.util.ClassUtil (file:/home/eldar/.cache/coursier/v1/https/repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.11.4/jackson-databind-2.11.4.jar) to field java.lang.Throwable.detailMessage
WARNING: Please consider reporting this to the maintainers of com.fasterxml.jackson.databind.util.ClassUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release`

PR #141, opened 2021-11-26
PersistentActor.ACKError should be undergo serialization verification

EldarKZ16:
UltimateSoftware/surge-ukg#103

### DIFF
--- a/modules/command-engine/core/src/main/scala/surge/exceptions/KafkaPublishTimeoutException.scala
+++ b/modules/command-engine/core/src/main/scala/surge/exceptions/KafkaPublishTimeoutException.scala
@@ -2,5 +2,5 @@
 
 package surge.exceptions
 
-case class KafkaPublishTimeoutException(aggregateId: String, underlyingException: Throwable)
-    extends RuntimeException(s"Aggregate $aggregateId timed out trying to publish to Kafka")
+case class KafkaPublishTimeoutException(aggregateId: String, cause: Throwable)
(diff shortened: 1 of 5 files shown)


modules/command-engine/core/src/main/scala/surge/exceptions/KafkaPublishTimeoutException.scala
```diff
@@ -3,4 +3,4 @@
 package surge.exceptions
 
 case class KafkaPublishTimeoutException(aggregateId: String, underlyingException: Throwable)
-    extends RuntimeException(s"Aggregate $aggregateId timed out trying to publish to Kafka", underlyingException)
+    extends RuntimeException(s"Aggregate $aggregateId timed out trying to publish to Kafka")
```
EldarKZ16, line 6: During the testing, I've found that nested exceptions can't be serialized/deserialized. That's why I've removed it.

jeffboutotte: Is that just for our exceptions? Or any exceptions (i.e. if the domain code throws a nested exception)? It seems unfortunate we'd lose all that additional context since the actual root cause is usually that `underlyingException`

EldarKZ16: I did some testing. And found that the issue was the **underlyingException** field in KafkaPublishTimeoutException. Renamed it to the **cause**.
Now I think it should be good.

jeffboutotte (approved):
Awesome - thank you!

PR #143, opened 2021-12-03
local actor provider issue fix

satyender071:
Added the config for the local actor issue.

In the below screenshots when the **config is set to local** the application log is giving the host and port as **localhost:3467** which I have changed as per the remote akka host and port that is  `**127.0.1.1:25520**`.


satyender071:
1. **Akka remote config**

akkaremoteConfig

2. **Application log when config set to remote**

remoteLog

3. **Akka local config**

AkkaLocalConfig

4. **Application logs** when config set to **local**

akkaLocalLog

jeffboutotte (approved):

PR #144, opened 2021-12-07
Mockito Test Helpers DSL

sebastianharko:
PR to be followed by another PR updating the docs


jeffboutotte (approved):
LGTM. At some point we can probably extract a test helper module to keep the artifact separate from the main engine but it's probably a bit early for that

PR #145, opened 2021-12-13
Fix SurgeMessagePipelineSpec; Improve RouterActor Testing

EldarKZ16:
https://github.com/UltimateSoftware/surge-ukg/issues/294
https://github.com/UltimateSoftware/surge-ukg/issues/175

### DIFF
--- a/modules/command-engine/core/src/main/scala/surge/core/SurgePartitionRouter.scala
+++ b/modules/command-engine/core/src/main/scala/surge/core/SurgePartitionRouter.scala
@@ -10,7 +10,7 @@ import surge.internal.SurgeModel
 import surge.internal.akka.kafka.KafkaConsumerPartitionAssignmentTracker
 import surge.internal.core.SurgePartitionRouterImpl
 import surge.internal.persistence.BusinessLogic
-import surge.kafka.PersistentActorRegionCreator
+import surge.kafka.{ KafkaProducerTrait, PersistentActorRegionCreator }
 import surge.kafka.streams._
 
 trait SurgePartitionRouter extends HealthyComponent {
@@ -26,7 +26,17 @@ object SurgePartitionRouter {
       kafkaStreamsCommand: AggregateStateStoreKafkaStreams[JsValue],
       regionCreator: PersistentActorRegionCreator[String],
       signalBus: HealthSignalBusTrait,
-      isAkkaClusterEnabled: Boolean): SurgePartitionRouter = {
-    new SurgePartitionRouterImpl(config, system, partitionTracker, businessLogic, kafkaStreamsCommand, regionCreator, signalBus, isAkkaClusterEnabled)
+      isAkkaClusterEnabled: Boolean,
+      kafkaProducerOverride: Option[KafkaProducerTrait[String, Array[Byte]]] = None): SurgePartitionRouter = {
+    new SurgePartitionRouterImpl(
+      config,
+      system,
+      partitionTracker,
+      businessLogic,
+      kafkaStreamsCommand,
+      regionCreator,
+      signalBus,
+      isAkkaClusterEnabled,
+      kafkaProducerOverride)
   }
 }
--- a/modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
+++ b/modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
@@ -3,11 +3,13 @@
 package surge.internal.core
 
 import akka.actor._
+import akka.cluster.{ Cluster, MemberStatus }
 import akka.cluster.sharding.external.ExternalShardAllocationStrategy
 import akka.cluster.sharding.{ ClusterSharding, ClusterShardingSettings }
 import akka.kafka.ConsumerSettings
 import akka.kafka.cluster.sharding.KafkaClusterSharding
 import akka.pattern.ask
+import akka.util.Timeout
 import com.typesafe.config.Config
 import org.apache.kafka.clients.consumer.ConsumerConfig
 import org.apache.kafka.common.serialization.{ ByteArrayDeserializer, StringDeserializer }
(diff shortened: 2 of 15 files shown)


modules/command-engine/core/src/main/scala/surge/core/SurgePartitionRouter.scala
```diff
@@ -26,7 +26,17 @@ object SurgePartitionRouter {
       kafkaStreamsCommand: AggregateStateStoreKafkaStreams[JsValue],
       regionCreator: PersistentActorRegionCreator[String],
       signalBus: HealthSignalBusTrait,
-      isAkkaClusterEnabled: Boolean): SurgePartitionRouter = {
-    new SurgePartitionRouterImpl(config, system, partitionTracker, businessLogic, kafkaStreamsCommand, regionCreator, signalBus, isAkkaClusterEnabled)
+      isAkkaClusterEnabled: Boolean,
+      kafkaProducerOverride: Option[KafkaProducerTrait[String, Array[Byte]]] = None): SurgePartitionRouter = {
```
EldarKZ16, line 30: I wasn't able to mock kafka producer for multi-jvm test. That's why this producer parameter has been introduced.

modules/command-engine/core/src/test/scala/surge/internal/domain/SurgeMessagePipelineSpec.scala
```diff
@@ -77,277 +64,240 @@ class SurgeMessagePipelineSpec
...
+    // FIXME verifySystemShutdown should be true, but this does not shut down in a reasonable amount of time
+    TestKit.shutdownActorSystem(system, duration = 30.seconds, verifySystemShutdown = false)
+    super.afterAll()
+  }
 
-            registrationNames should contain("state-store-kafka-streams")
-            val kStreamsRegistration = registrations.find(r => r.componentName == "state-store-kafka-streams").get
-            kStreamsRegistration.restartSignalPatterns.map(p => p.pattern()) should contain("kafka.streams.fatal.error")
+  val ctx: TestContext = createTestContext()
+  val pipeline: SurgeMessagePipeline[State, BaseTestCommand, BaseTestEvent] = ctx.pipeline
```
EldarKZ16, line 144: The class initializes a single SurgeMessagePipeline. 
Because we can't start and stop the pipeline in a fast manner between the tests and it causes Actor name is not unique issue.

modules/command-engine/core/src/multi-jvm/scala/surge/internal/core/SurgePartitionRouterImplMultiJvmSpec.scala
```diff
@@ -0,0 +1,212 @@
...
+    val cqrsRegionCreator = new ProbeInterceptorRegionCreator(regionProbe)
+
+    val producer = mock[KafkaProducerTrait[String, Array[Byte]]]
+    when(producer.topic).thenReturn(trackedTopic)
+    when(producer.partitionFor(anyString)).thenAnswer((invocation: InvocationOnMock) => {
+      val key = invocation.getArgument[String](0)
+      partitionMappings.get(key)
+    })
+
+    lazy val actorRouter: SurgePartitionRouterImpl =
```
jeffboutotte, line 107: lazy here seems unnecessary since it's used on the line below

modules/common/src/multi-jvm/scala/surge/kafka/KafkaPartitionShardRouterActorMultiJvmSpec.scala
```diff
@@ -91,5 +95,29 @@ class KafkaPartitionShardRouterActorSpecBase
         probe.expectMsg(command0)
       }
     }
+
+    "Forward message and fail" in {
```
jeffboutotte, line 99: What's the value in this test? It demonstrates the bug we fixed awhile back but I don't think it catches any regressions in the actual code around that?

modules/command-engine/core/src/main/scala/surge/internal/domain/SurgeMessagePipeline.scala
```diff
@@ -70,7 +70,7 @@ private[surge] abstract class SurgeMessagePipeline[S, M, E](
   protected val cqrsRegionCreator: PersistentActorRegionCreator[M] =
     new PersistentActorRegionCreator[M](actorSystem, businessLogic, kafkaStreamsImpl, partitionTracker, businessLogic.metrics, signalBus, config = config)
 
-  protected val actorRouter: SurgePartitionRouter =
+  protected lazy val actorRouter: SurgePartitionRouter =
```
jeffboutotte, line 73: Any reason in particular to make this lazy?

modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
```diff
@@ -60,11 +66,17 @@ private[surge] final class SurgePartitionRouterImpl(
...
       businessLogic.kafka.stateTopic,
       regionCreator,
-      RoutableMessage.extractEntityId)(businessLogic.tracer)
+      RoutableMessage.extractEntityId,
+      kafkaProducerOverride)(businessLogic.tracer)
 
     val routerActorName = s"${businessLogic.aggregateName}RouterActor"
-    val shardRouter = system.actorOf(shardRouterProps, name = routerActorName)
-    shardRouter
+    val shardRouterManaged = ActorLifecycleManagerActor.manage(
```
jeffboutotte, line 73: Let's leave this not wrapped in a managed actor for now if possible. Given the subtlety of the bug around this that was fixed and the fact that the router actor goes away when we swap to using Akka Cluster I'm a bit anxious that the tradeoff of being able to control the actor in this way isn't worth the risk of some other subtle change coming along and completely breaking multi-node down the line

modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
```diff
@@ -60,11 +66,17 @@ private[surge] final class SurgePartitionRouterImpl(
...
+      actorSystem = system,
+      managedActorProps = shardRouterProps,
+      componentName = routerActorName,
+      actorLifecycleName = Some(routerActorName),
```
jeffboutotte, line 77: Actually one thing you could do to validate there are sufficient tests is remove the `actorLifecycleName` parameter (or pass in a `None`) and validate that we have a failing test that was the bug around this. Having a reliably failing test around that would ease my anxiety about re-wrapping this in a managed actor

modules/command-engine/core/src/main/scala/surge/internal/domain/SurgeMessagePipeline.scala
EldarKZ16: In the SurgeMessagePipelineSpec, when we override the actorRouter it is evaluated twice. Lazy val resolves that issue.

modules/command-engine/core/src/main/scala/surge/internal/core/SurgePartitionRouterImpl.scala
EldarKZ16: Ok, agree with you. Thanks! I'll remove it completely

modules/common/src/multi-jvm/scala/surge/kafka/KafkaPartitionShardRouterActorMultiJvmSpec.scala
EldarKZ16: Initially, I've tried to reproduce it locally, but it works correctly(in my case all actor lifecycle name was the same for multiple nodes). So I've added this test, just to demonstrate the bug as you said

jeffboutotte: Oh gotcha, the original bug was fixed but we just needed something in place to help catch regressions. I think we can get rid of this test here since it's not actually testing for the regression, just demonstrating the original bug

jeffboutotte (approved):
LGTM - thanks!

EldarKZ16:
Oops, sorry for this commit.

jeffboutotte (approved):

PR #147, opened 2021-12-14
DRAFT PR metrics integration KTable observability

satyender071:
Added couple of metrics to `AggregateStateStoreKafkaStreamsImpl`.


modules/common/src/main/scala/surge/kafka/streams/AggregateStateStoreKafkaStreamsImpl.scala
```diff
@@ -55,6 +56,24 @@ private[streams] class AggregateStateStoreKafkaStreamsImpl[Agg >: Null](
...
+      tags = Map("aggregate" -> aggregateName)))
+
+  private val getAggregateBytesTimerMetric: Timer = metrics.timer(
+    MetricInfo(
+      name = s"surge.${aggregateName.toLowerCase()}.get-aggregate-state",
+      description = "The time taken to getting the aggregate state.",
+      tags = Map("aggregate" -> aggregateName)))
+
+  val current: Rate = metrics.rate(
+    MetricInfo(
```
jeffboutotte, line 72: I think we've already got this metric on the producer actor somewhere - this is capturing rates on the KTable read side but I'd be curious if we can capture metrics on the KTable write side. It may be that Kafka Streams already has some metrics for that, but I haven't looked too closely. Unfortunately in this case (but good for other reasons) our stream topology is so simple that there's nowhere really to put a custom metric for rate on the input side

satyender071: Yeah sure i will look for the same meanwhile

jeffboutotte:
This should be covered by #179 - it pulled that metric in for getting aggregate state and a previous change gets rid of the getSubstatesForAggregate piece since it was never really exposed through the APIs

PR #150, opened 2021-12-17
Retry mechanism in javadsl when engine is still starting

rohan-sircar:
Add a new state called `Starting` in SurgeEngineStatus. The engine's start method sets the state to `Starting` immediately, and sets to `Running` at the end.

In javadsl's AggregateRef, sendCommand and getState methods wait until engine is no longer in starting state in a retry loop for `maxRetries` iterations.

This is done so users of javadsl do not have to flatMap on the future returned from start.


modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/AggregateRef.scala
```diff
@@ -38,31 +41,51 @@ final class AggregateRefImpl[AggId, Agg, Cmd, Event](
...
-    } else {
-      log.error(s"Engine Status: $status")
-      Future.failed(SurgeEngineNotRunningException("The engine is not running, please call .start() on the engine before interacting with it"))
     }
 
-    FutureConverters.toJava(result)
+    FutureConverters.toJava(r)
   }
+
+  private def retryWithDelay(numTries: Int, delay: FiniteDuration = 50.millis): Future[Unit] = {
```
sebastianharko, line 78: can we rename this to "whenEngineReady"

sebastianharko: How about something like this : 

```
whenEngineReady[T](numTries: Int = 5, delay: FiniteDuration = 50.millis)(codeBlock: _ => Future[T]): Future[T] = {

}
```

rohan-sircar: Done, but I kept codeblock to ` cb: => Future[T]`

sebastianharko: Amazing ! Thanks ;-)

jeffboutotte (approved):
This makes sense and is an improvement over the way things currently are, but wondering if this retry mechanism is just masking over some underlying problem - perhaps the engine should be stashing messages while starting? Or maybe it just needs certain components to be up (maybe even just the router which may stash things?) before we consider the engine running

PR #189, opened 2021-12-24
Fix EmbeddedKafka and SurgeMessagePipelineSpec test

EldarKZ16:


modules/command-engine/core/src/test/scala/surge/internal/domain/SurgeMessagePipelineSpec.scala
```diff
@@ -132,9 +132,10 @@ class SurgeMessagePipelineSpec
   }
 
   override def afterAll(): Unit = {
-    pipeline.controllable.stop().futureValue shouldBe an[Ack]
-    TestKit.shutdownActorSystem(system, verifySystemShutdown = true)
     EmbeddedKafka.stop()
+    pipeline.controllable.stop().futureValue shouldBe an[Ack]
```
jeffboutotte, line 136: I think there's a race condition here where if Kafka shuts down before the engine it doesn't always stop cleanly in a reasonable amount of time which is why `verifySystemShutdown = true` wasn't quite working originally, not sure about why its flaky now though

modules/multilanguage/src/test/resources/application.conf
```diff
@@ -1,6 +1,7 @@
-kafka.port = 2021
+kafka.port = 57456
```
jeffboutotte, line 1: Can these be a random open port instead of a hard coded one? I know we need to pass it to the engine, but the config for that should be able to be passed through hopefully.

project/Dependencies.scala
```diff
@@ -33,7 +33,7 @@ object Dependencies extends AutoPlugin {
     }
 
     object Kafka {
-      val kafkaVersion = "2.7.2"
+      val kafkaVersion = "2.8.1"
```
jeffboutotte, line 36: I don't think we're quite ready for this upgrade yet, the update to 2.7 was a bit bumpier than expected even though it should be backwards compatible and generally smooth

jeffboutotte (commented):
This looks fine overall, but how do we know these are actually stable? We've disabled/enabled these tests (particularly the `SurgeMessagePipelineSpec`) several times now. In general are these tests just doing too much maybe? Given the flakiness of them maybe the level of testing we get from `SurgeMessagePipelineSpec` isn't worth the tests being unreliable?

project/Dependencies.scala
EldarKZ16: Agree, reverted this one.

modules/multilanguage/src/test/resources/application.conf
EldarKZ16: Yeah, it can be any other random open port.

modules/command-engine/core/src/test/scala/surge/internal/domain/SurgeMessagePipelineSpec.scala
EldarKZ16: Also not sure why it's flaky. I've tried to revert changes and now seeing a bunch of errors.
https://github.com/UltimateSoftware/surge/pull/189/commits/d9307fe5c65ff906aaab79af30046883ef7384a2

PR #194, opened 2021-12-28
add uninitialized receive to HostAssignmentTrackerImpl

EldarKZ16:
Issue: https://github.com/UltimateSoftware/surge-ukg/issues/352
When the app is initialized, it takes some time for Kafka host assignment. So, HostAssignmentTracker should not receive GetState messages before kafka initialization.


modules/common/src/main/scala/surge/internal/kafka/HostAssignmentTracker.scala
```diff
@@ -52,8 +52,13 @@ object HostAssignmentTracker {
   private case object ClearState
 
   private case class ClusterState(state: Map[TopicPartition, HostPort])
-  private class HostAssignmentTrackerImpl extends Actor {
-    override def receive: Receive = receiveWithState(ClusterState(Map.empty))
+  private class HostAssignmentTrackerImpl extends Actor with Stash {
+    override def receive: Receive = uninitialized()
+
+    private def uninitialized(): Receive = {
+      case msg: UpdateState => handleUpdateState(ClusterState(Map.empty), msg)
```
jeffboutotte, line 59: Should we just `unstashAll()` here rather than in `handleUpdateState` since we're only ever stashing messages from the `uninitialized` state?

jeffboutotte (approved):

PR #204, opened 2022-01-10
track published messages

zachary-rote:


modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -161,24 +192,62 @@ class KafkaProducerActor(
...
     aggregateName: String,
     val assignedPartition: TopicPartition,
-    override val signalBus: HealthSignalBusTrait)
+    override val signalBus: HealthSignalBusTrait,
+    trackerTimeout: FiniteDuration = TimeoutConfig.PublisherActor.trackerTimeout)
     extends HealthyComponent
     with HealthSignalBusAware {
   private implicit val executionContext: ExecutionContext = ExecutionContext.global
   private val log = LoggerFactory.getLogger(getClass)
+  private val inflight: mutable.Set[MessageTracker] = mutable.Set.empty
```
jeffboutotte, line 201: I think the set of tracked messages wants to live in the actor for thread safety - outside of the actor there's no guarantee two things won't try to mutate this set at the same time and step on one another, but within the actor we're guaranteed to process one message at a time and avoid that race condition.

That might mean a few other things will need to move around as well - `KTablePersistenceSupport` will no longer have access to the tracker directly if it lives in the actor state, so it can just blindly retry and we can let the smarter producer actor sort out duplicates

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -347,86 +401,97 @@ class KafkaProducerActorImpl(
...
-            new ProducerRecord(eventsTopic.name, null, eventToPublish.key, eventToPublish.value, eventToPublish.headers) // scalastyle:ignore null
-          })
-        .getOrElse(Seq.empty)
-
-      val stateRecords = stateMessages.map { state =>
-        new ProducerRecord(stateTopic.name, assignedPartition.partition(), state.key, state.value, state.headers)
-      }
-      val records = eventRecords ++ stateRecords
+      // Writes grouped by requestId
+      val groupedWrites = state.groupedWrites()
```
jeffboutotte, line 405: Hmm this groups by requestId here? So 2 different aggregates publishing would end up with 2 different request ids which would get grouped into 2 different groups? If that's the case I think this might be removing the grouping optimization we have here for transactions where we lump publish requests from multiple aggregates into one Kafka transaction to reduce overhead (one transaction per aggregate is slow and expensive).

To get around that I think instead of keeping a list of senders it could be a list of sender + requestId that we track (here are all the request ids that are in flight along with the sender who's waiting for a response)

garyKeorkunian (approved):

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
```diff
@@ -46,7 +50,16 @@ trait KTablePersistenceSupport[Agg, Event] {
       serializedState: KafkaProducerActor.MessageToPublish,
       startTime: Instant)
       extends Internal
-  private case class EventPublishTimedOut(reason: Throwable, startTime: Instant) extends Internal
+  private case class EventPublishTimedOut(
+      trackingId: UUID,
```
jeffboutotte, line 54: I think this needs to track a `newState` in this object since that's eventually used in the case of persistence success, the `state` passed into `handleInternal` is the old state (in case of failure)

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
```diff
@@ -62,30 +75,43 @@ trait KTablePersistenceSupport[Agg, Event] {
...
+          case KafkaProducerActor.PublishSuccess(publishRequestId) => PersistenceSuccess(publishRequestId, state, startTime, context)
+          case KafkaProducerActor.PublishFailure(publishRequestId, t, _) =>
+            PersistenceFailure(publishRequestId, state, context, t, currentFailureCount + 1, serializedEvents, serializedState, startTime)
         }
         .recover { case t =>
           log.error("Failed to publish messages", t)
-          EventPublishTimedOut(t, startTime)
+
+          t match {
+            case exception: RetryAwareException =>
```
jeffboutotte, line 95: I think we should track `currentFailureCount` here as well to put a ceiling on how many attempts before we give up

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -225,18 +249,44 @@ class KafkaProducerActorImpl(
...
+                }
+              } else {
+                context.become(processing(state.stopTracking(msg)))
+                context.self ! EventsFailedToPublish(
+                  reason = new RuntimeException(s"PublishTracker expired, not attempting another publish for ${msg.batchId}"),
+                  originalSenders = Seq(SenderWithTrackingId(sender(), msg.batchId)))
+              }
+            } else {
+              context.become(processing(state))
+              context.self ! EventsPublished(Seq(SenderWithTrackingId(sender(), msg.batchId)))
```
jeffboutotte, line 271: We can just directly reply with success in this case rather than sending a message to ourself to reply successful to avoid an extra step/unnecessary message in the mailbox to process

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -123,9 +126,43 @@ object KafkaProducerActor {
...
 
   sealed trait PublishResult extends NoSerializationVerificationNeeded
-  case object PublishSuccess extends PublishResult
-  case class PublishFailure(t: Throwable) extends PublishResult
+  case class PublishSuccess(publishRequestId: UUID) extends PublishResult
+  case class PublishFailure(publishRequestId: UUID, t: Throwable, retry: Boolean = false) extends PublishResult
   case class MessageToPublish(key: String, value: Array[Byte], headers: Headers)
+
+  object PublishTracker {
+    def apply(publishTrackingId: UUID, data: (MessageToPublish, Seq[MessageToPublish]), trackerTimeout: FiniteDuration): PublishTracker = {
```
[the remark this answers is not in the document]
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/core/KafkaProducerActor.scala
```diff
@@ -123,9 +126,43 @@ object KafkaProducerActor {
...
+      new SurgePublishTracker(publishTrackingId, state = Some(data._1), events = data._2, dataWasPublished = false, ttl = trackerTimeout)
+    }
+
+    def alreadyPublished(publishTrackingId: UUID, trackerTimeout: FiniteDuration): PublishTracker = {
+      new SurgePublishTracker(publishTrackingId, state = None, events = Seq.empty, dataWasPublished = true, ttl = trackerTimeout)
+    }
+  }
+
+  sealed trait PublishTracker {
```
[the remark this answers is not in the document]
zachary-rote: done

modules/common/src/main/resources/reference.conf
```diff
@@ -170,6 +170,9 @@ surge {
     publish-timeout = 15 seconds
     publish-timeout = ${?SURGE_PRODUCER_PUBLISH_TIMEOUT}
 
+    tracker-timeout = 60 seconds
+    tracker-timeout = ${?SURGE_PRODUCER_TRACKER_TIMEOUT}
```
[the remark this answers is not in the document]
jeffboutotte: Aah actually after looking at some of the other logic I think in step 6 there we would have id = 1 in the state as tracked so we wouldn't add to the pending writes and would reply...but that may be a different issue then. Tracked items are expired in the failure case - they're expired and the caller retries on an expired request but are never expired in the successful case which could lead to the state getting massive over time if everything publishes successfully (which it ideally should)

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -225,18 +249,44 @@ class KafkaProducerActorImpl(
...
+        val maybeTrackerWithExpiry = publishTrackerStateManager.tracker(msg.batchId, state)
+        maybeTrackerWithExpiry match {
+          case Some(trackerWithExpiry) =>
+            if (!trackerWithExpiry.tracker.dataWasPublished()) {
+              if (!trackerWithExpiry.expired) {
+                // Avoid duplicating a write that is already pending.
+                if (!publishTrackerStateManager.stateHasPendingWrites(msg.batchId, state)) {
+                  context.become(processing(state.addPendingWrites(sender(), msg)))
+                } else {
+                  context.become(processing(state))
```
jeffboutotte, line 261: In this case (if I follow correctly) data hasn't been published but we're already tracking things - i.e. there is a current write to Kafka in progress. Wouldn't we want to update the original sender for the message so that the new asker is notified of completion rather than the old one (who has since given up)?

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -317,15 +367,17 @@ class KafkaProducerActorImpl(
   }
 
   private def handle(state: KafkaProducerActorState, eventsPublished: EventsPublished): Unit = {
-    val newState = state.addInFlight(eventsPublished.recordMetadata).completeTransaction()
-    context.become(processing(newState))
-    eventsPublished.originalSenders.foreach(_ ! KafkaProducerActor.PublishSuccess)
+    if (eventsPublished.recordMetadata.nonEmpty) {
```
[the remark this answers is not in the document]
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -32,7 +36,20 @@ object KafkaProducerActorImpl {
   val KAFKA_PRODUCER_FENCED_SIGNAL_NAME: String = "kafka.producer.actor.fenced"
 
   sealed trait KafkaProducerActorMessage extends NoSerializationVerificationNeeded
-  case class Publish(state: KafkaProducerActor.MessageToPublish, eventsToPublish: Seq[KafkaProducerActor.MessageToPublish]) extends KafkaProducerActorMessage
+
+  class PublishTrackerStateManager(val trackerTimeout: FiniteDuration = TimeoutConfig.PublisherActor.trackerTimeout) {
```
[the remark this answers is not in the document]
zachary-rote: used to mock inaccessible state in spec.

modules/common/src/main/resources/reference.conf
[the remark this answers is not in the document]
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
```diff
@@ -62,30 +77,69 @@ trait KTablePersistenceSupport[Agg, Event] {
...
+                trackingId = trackingId,
+                aggregateId = aggregateId,
+                context = context,
+                state = serializedState,
+                events = serializedEvents,
+                reason = t,
+                startTime = startTime,
+                retry = exception.retry,
+                newState = state,
+                numberOfFailures = currentFailureCount)
```
jeffboutotte, line 108:
```suggestion
                numberOfFailures = currentFailureCount + 1)
```

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
```diff
@@ -62,30 +77,69 @@ trait KTablePersistenceSupport[Agg, Event] {
...
+              EventPublishTimedOut(
+                trackingId = trackingId,
+                aggregateId = aggregateId,
+                context = context,
+                state = serializedState,
+                events = serializedEvents,
+                reason = t,
+                startTime = startTime,
+                newState = state,
+                numberOfFailures = currentFailureCount)
```
jeffboutotte, line 119:
```suggestion
                numberOfFailures = currentFailureCount + 1)
```

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -122,6 +147,9 @@ class KafkaProducerActorImpl(
   private implicit val rates: AggregateStateRates = AggregateStateRates(aggregateName, metrics)
 
   context.system.scheduler.scheduleOnce(10.milliseconds, self, InitTransactions)
+
+  context.system.scheduler.scheduleAtFixedRate(10.milliseconds, 10.seconds, self, ClearExpiredTrackers)
```
jeffboutotte, line 151: Can you cancel this task on shutdown as well? Similar to `flushMessagesScheduledTask` below

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
```diff
@@ -203,9 +231,11 @@ class KafkaProducerActorImpl(
     case msg: ProducerFenced => maybeRestartFencedProducer(msg)
     case msg: EventsFailedToPublish =>
       context.become(fenced(state.completeTransaction(), initialFenceTime))
-      msg.originalSenders.foreach(_ ! KafkaProducerActor.PublishFailure(msg.reason))
-    case RestartProducer              => restartPublisher()
-    case ShutdownProducer             => stopPublisher()
+      msg.originalSenders.foreach(sender => sender.actorRef ! KafkaProducerActor.PublishFailure(sender.trackingId, msg.reason))
+    case RestartProducer  => restartPublisher()
+    case ShutdownProducer => stopPublisher()
+    case ClearExpiredTrackers =>
```
jeffboutotte, line 237: We should handle this in the other states for this actor as well to avoid dead letters from the scheduled message not being handled since we're kinda expecting them to show up all the time in any state. Handling can just be ignoring it:
```
case ClearExpiredTrackers => // Ignore and do nothing
```

jeffboutotte (commented):
Couple minor things on the scheduled task + I think the failure count needs to be incremented to track correctly (unless I'm missing somewhere else it's incremented)

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/kafka/KafkaProducerActorImpl.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
zachary-rote: done

modules/command-engine/core/src/main/scala/surge/internal/persistence/KTablePersistenceSupport.scala
zachary-rote: done

PR #235, opened 2022-03-23
Persistence Interface First Attempt / Proposal

sebastianharko:


modules/common/src/main/scala/surge/core/SurgePersistence.scala
```diff
@@ -0,0 +1,12 @@
+package surge.core
+
+import akka.Done
+
+import scala.concurrent.Future
+
+trait SurgePersistence[State, Event] {
```
jeffboutotte, line 7: I think it's worth splitting this a little early since we know eventually we'll have different sorts of entities - a CQRS one (current command model) and a stateful streaming one (current event model) and potentially even more custom things. I had this snippet stashed away with my thinking a little while back but am happy to adjust and go with what makes the most sense
```
trait SurgeEntityContext[State] {
  // Will need some serialization bits here
  def persistState(state: Option[State]): Future[Done]
  def fetchState(id: String): Future[Option[State]]
}

trait SurgeCQRSEntityContext[State, Event] extends SurgeEntityContext[State] {
  def persistEvents(events: Seq[Event]): Future[Done]
  def persistAll(events: Seq[Event], state: Option[State]): Future[Done] // Maybe just pass serialized things here instead??
}
```

sebastianharko: Regarding the name SurgeCQRSEntityContext ..can we change it to something else ... CQRS in an architecture principle and can be implemented even without event sourcing, so I propose the following names: 1. SurgeSnapshotting 2.SurgeEventPublishing ?? Let me know what you think ...

jeffboutotte: I'm good with that naming!

PR #237, opened 2022-03-24
Add readiness healthcheck to surge engine

rohan-sircar:
for https://github.com/UltimateSoftware/surge-ukg/pull/473

Adds a readiness healthcheck to surge engine, as a method `def ready(): Future[HealthCheck]` in the HealthCheckTrait trait,
and it delegates to akka management, without starting an AM http server.
Liveness check is already implemented in the existing `healthCheck()` method.

### DIFF
--- a/modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/SurgeCommand.scala
+++ b/modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/SurgeCommand.scala
@@ -84,6 +84,10 @@ private[javadsl] class SurgeCommandImpl[AggId, Agg, Command, Evt](
     FutureConverters.toJava(healthCheck().map(_.asJava))
   }
 
+  def getReady: CompletionStage[HealthCheck] = {
+    FutureConverters.toJava(ready().map(_.asJava))
+  }
+
   def aggregateFor(aggregateId: AggId): AggregateRef[Agg, Command, Evt] = {
     new AggregateRefImpl(aggIdToString(aggregateId), actorRouter.actorRegion, businessLogic.tracer, getEngineStatus)
   }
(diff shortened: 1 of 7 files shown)


modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/SurgeCommand.scala
```diff
@@ -84,6 +84,10 @@ private[javadsl] class SurgeCommandImpl[AggId, Agg, Command, Evt](
     FutureConverters.toJava(healthCheck().map(_.asJava))
   }
 
+  def getReady: CompletionStage[HealthCheck] = {
```
jeffboutotte, line 87: Maybe `getReadiness` for naming on this?

modules/command-engine/scaladsl/src/main/scala/surge/scaladsl/common/HealthCheckTrait.scala
```diff
@@ -9,4 +9,5 @@ import scala.concurrent.Future
 
 trait HealthCheckTrait {
   def healthCheck: Future[HealthCheck]
+  def ready: Future[HealthCheck]
```
jeffboutotte, line 12: Maybe `readiness` on this as well for naming

rohan-sircar: done

modules/command-engine/javadsl/src/main/scala/surge/javadsl/command/SurgeCommand.scala
rohan-sircar: done

rohan-sircar:
@jeffboutotte the most recent version of akka management still depends on akka 2.6.14 while surge expects 2.6.15. There's several ways we could go about this.

jeffboutotte:
Are we actually leveraging management for anything at the moment besides these new readiness checks? We could pull in 2.6.15 of whatever akka management needs to keep everything on a consistent version, but wondering what the value we're getting from all the extra things we pull along is.

PR #270, opened 2022-07-18
fix ackerror serialization/deserialization

zachary-rote:


modules/command-engine/core/src/test/scala/surge/internal/persistence/PersistentActorSpec.scala
```diff
@@ -171,6 +175,18 @@ class PersistentActorSpec
   }
 
   "PersistentActor" should {
+    "Serialize and Deserialize AckError" in {
+      val error = ACKError(new Exception("test-failure"))
```
jeffboutotte, line 179: You can test this through the Akka serializer as well (which is what actually gets used). We have a few of those in this suite actually. Looking a little closer it looks like maybe the [`compareStringResults` bit](https://github.com/UltimateSoftware/surge/blob/4b483d9eeb2a97c9a81b275e4ee7c3a4314d1919/modules/command-engine/core/src/test/scala/surge/internal/persistence/PersistentActorSpec.scala#L629-L633) from those might be hiding some issues with serialization

zachary-rote: done

jeffboutotte (approved):

